PR #2, opened 2013-03-07
fix CRM-12047

dlobo:


totten:
Dudes! Notice the first character of the first change in 71bafb2.

dlobo:
i think kurund fixed this. we should extend the pre-hook code to also 
cover .module and .inc files in it php syntax check

lobo

kurund:
yes, I have fixed the syntax, error.

PR #14, opened 2013-04-30
CRM-12476

dlobo:
---
- CRM-12476: CRM_Contribute_PseudoConstant  wrong method contributionType not fiancialType
  http://issues.civicrm.org/jira/browse/CRM-12476


dlobo:
oops should do this on 4.3

PR #16, opened 2013-05-01
Refactor v2 api calls CRM-9794

colemanw:
---
- CRM-9794: Remove API v2
  http://issues.civicrm.org/jira/browse/CRM-9794


eileenmcnaughton:
Hi, are you applying these same fixes to the d6 branch too?

colemanw:
Oh you're so accommodating Eileen

eileenmcnaughton:
well - breaking d6 code for the sake of removing cruft doesn't make sense to me... (while d6 people might be lessening in number my experience is that they have been among the more significant code / patch / MIH contributors - at least in our stable)

colemanw:
See https://github.com/civicrm/civicrm-drupal/pull/24

PR #18, opened 2013-05-10
civicrm.drush.inc - Set current user before executing civicrm_api

totten:


totten:
This allows one to set a user's API key via drush, e.g.

```
drush -u <username> civicrm-api contact.create id=user_contact_id api_key=<api-key>
```

See also: http://issues.civicrm.org/jira/browse/CRM-12556

totten:
Discussed on IRC

PR #24, opened 2013-05-23
Refactor v2 api calls CRM-9794

colemanw:
---
- CRM-9794:
  http://issues.civicrm.org/jira/browse/CRM-9794


eileenmcnaughton:
thanks

PR #29, opened 2013-05-28
worked on CRM-12708, added filter for multi-select state province

monishdeb:
---
- CRM-12708: Missing filter handler for Multi-Select State/Province
  http://issues.civicrm.org/jira/browse/CRM-12708


monishdeb:
Sorry to create the PR for master again

PR #44, opened 2013-08-02
CRM-12874

dlobo:
Renamed it to civicrm.settings.php.template

CRM-12874: Rename civicrm.settings.php.tpl so it isn't in the same suffix as Smarty templates http://issues.civicrm.org/jira/browse/CRM-12874


totten:
I think the drush/provision process tries to support multiple versions. Might do:

``` php
$files = array(
  "$crmPath/templates/CRM/common/civicrm.settings.php.template",
  "$crmPath/templates/CRM/common/civicrm.settings.php.tpl"
);
foreach ($files as $file) {
  if (file_exists($file)) {
    $settingsTplFile = $file;
  }
}
```

ping @mlutfy

dlobo:
fixed and recommitted

PR #45, opened 2013-08-03
CRM-12853 Case/Views integration

mcleinn:
@Kurund Jalmi 

Please see notes add:
http://issues.civicrm.org/jira/browse/CRM-12853

I removed the bugfix mentioned in Note 1, as the code has changed in 4.4, and the bug might have been removed.
The test links still work with 4.4 code:

 http://viazni.pakalennie.de/cases
 http://viazni.pakalennie.de/activities
 http://viazni.pakalennie.de/relationships

CiviCase also appears in the menus for setting up views, but I can't completely check all the configuration details again at the moment.

As an aside dlats@netzero.net wrote about my patch:
"Thanks again for the patch to _civicrm_core_data (activity_id). In looking through your code, I noticed that you specified joins to civicrm_activity and civicrm_contacts as well as adding a relationship to the fields civicrm_case_activity.activity_id and civicrm_case_contact.contact_id. I was able to use just the relationships with out specifying the joins. It seems redundant."

So it is quite possible that some code is redundant, however I am weary to remove anything, as it seems to work.

### DIFF
--- a/modules/views/civicrm.views.inc
+++ b/modules/views/civicrm.views.inc
@@ -76,6 +76,7 @@ function civicrm_views_data_alter(&$data) {
   require_once 'CRM/Core/Error.php';
   require_once 'CRM/Contact/BAO/Contact.php';
   require_once 'CRM/Event/BAO/Query.php';
+  require_once 'CRM/Case/BAO/Case.php';
   require_once 'components/civicrm.core.inc';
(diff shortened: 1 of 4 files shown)


kurund:
thanks for the patch, there are few indentation issues, which we will fix

PR #58, opened 2013-09-12
CRM-13388 - Fix upgrade error due to undefined column, "description"

totten:
---
- CRM-13388: Drush upgrade fails due to missing column, civicrm_uf_group.description
  http://issues.civicrm.org/jira/browse/CRM-13388


totten:
Depends on https://github.com/civicrm/civicrm-core/pull/1632

PR #62, opened 2013-09-30
CRM-13470: add hook_flush_caches

kreynen:
This is my first CiviCRM PR, so let me know if I'm not doing it right

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -1213,28 +1213,3 @@ function civicrm_modules_uninstalled($modules) {
   CRM_Core_Invoke::rebuildMenuAndCaches(TRUE);
 }
 
-/**
- * Implementation of hook_flush_caches().
- */
-function civicrm_clear_all_caches_flush_caches() {
-
-  if (!civicrm_initialize()) {
-    return;
-  }
-
-  require_once 'CRM/Core/Config.php';
-  $config = CRM_Core_Config::singleton();
-
-  // clear db caching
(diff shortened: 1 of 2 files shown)


totten:
1. Looking at the Github tab for "Files Changed", it seems that the net effect of this PR is to modify a label without actually adding the hook (ie f57b49c cancelsout 1a76b32). Is that intentional, accidental, or a Github bug?
2. The rebuildMenuAndCaches() function is the main logic for a "general system flush/reset". This can be reached via web (eg "/civicrm/menu/rebuild" or "/civicrm/clearcahe") or via API (eg "drush cvapi System.flush").
3. In other parts of civicrm.module which need to flush caches (like hook_modules_installed), it calls:

``` php
  civicrm_initialize();
  CRM_Core_Invoke::rebuildMenuAndCaches(TRUE);
```

If we used hook_flush_caches, could we get rid of the other calls? We need to make sure that the Civi caches are cleared whenever Drupal modules (or Joomla extensions or Civi extensions or whatever) are installed/disabled/enabled/uninstalled.

kreynen:
Discussed this w/ @totten on IRC.  We agreed that the hook_flush_caches should stay as a contrib module.  I will add the option of using the more thorough CRM_Core_Invoke::rebuildMenuAndCaches(TRUE) to https://drupal.org/project/civicrm_clear_all_caches

PR #64, opened 2013-10-07
Restore old hardcoded join to the relationship table for backwards compa...

torrance:
...tibility.


colemanw:
This was submitted against master, I've manually merged into 4.4

PR #68, opened 2013-10-11
Exposes all the fields in the civicrm_mailing table to drupal views

caprenter:
More or less copied and pasted the examples for other rows that are currently
exposed to views to add all the other rows in the table.

Some work could be done on the user experience for using some of the more
obscure fields, but it should give a views user the ability to use many
more fields.

Don't have a specific ticket to commit this against sorry!


colemanw:
Looks good. Thanks for the PR.

PR #72, opened 2013-10-21
Implicitly join civicrm_contributions to civicrm_contacts

torrance:
Also fixes bug in removed relationship, where the join was specified on the wrong fields.


kurund:
Can you please resend the PR against 4.4 branch and also file an issue for the same in jira

PR #73, opened 2013-10-21
Fix errors in civicrm_handler_field_money

torrance:


kurund:
Can you please resend the PR against 4.4 branch and also file an issue for the same in jira.

PR #89, opened 2013-11-12
CRM-13737 fix d6 code to facilite altered directory structure

eileenmcnaughton:


civicrm.module
```diff
@@ -1090,7 +1079,6 @@ function civicrm_filter($op, $delta = 0, $format = -1, $text = "") {
       civicrm_initialize();
       $config = CRM_Core_Config::singleton();
       $smarty = CRM_Core_Smarty::singleton();
-      require_once 'CRM/Core/Smarty/resources/String.php';
```
totten, line 1093: I think we need to keep this one. The others deal with straight-up classes, but the symbol being imported here is actually a function (which means the autoloader doesn't get involved).

eileenmcnaughton, line 1093: Yes - you are right - I put that back - argh - why are there always some things that are different

eileenmcnaughton, line 1093: Also - https://github.com/civicrm/civicrm-drupal/pull/90/files to re-add for d7

PR #94, opened 2013-12-01
Set views field handlers for gender and phone type to use new pseudoconstant access.

torrance:


colemanw:
torrance can you change 'DAO' to 'BAO' in the 1st param to ::get
thanks

torrance:
Updated to BAO.

PR #96, opened 2013-12-03
Big cleanup of civicrm.views.inc

torrance:
An attempt to clean up civicrm.views.inc, in particular to make the function civicrm_views_custom_data_cache() more easier to follow.

This shouldn't change any behaviour.


dlobo:
torrance:

can u issue this PR against 7.x-4.4

thanx

PR #110, opened 2014-01-20
CRM-14096 civicrm.config code doesn't do a good job when civicrm not in ...

eileenmcnaughton:
...sites/all

### DIFF
--- a/civicrm.config.php.drupal
+++ b/civicrm.config.php.drupal
@@ -43,16 +43,6 @@ function civicrm_conf_init() {
     }
 
     /**
-     * Quick test to see if the config file is in the standalone directory
-     */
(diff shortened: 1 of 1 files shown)


civicrm.config.php.drupal
```diff
@@ -86,7 +85,7 @@ function civicrm_conf_init() {
         echo "<a href=\"../install/index.php?mode=standalone\">click here</a>.\n";
```
eileenmcnaughton: do we still need this standalone comment?

totten: ^^ @cividesk

eileenmcnaughton:
Hang fire on this - reviewing

eileenmcnaughton:
OK - this is good to go - the file is now exactly the same as the d7 version of the file - which had improvements not ported back to d6, I added one minor tweak onto d7 too

PR #118, opened 2014-02-07
CRM-14186: drush cache clear: use the API to flush the code, and slip…

mlutfy:
in a fix to delete the civicrm_class_loader drupal variable that can cause problems if the civicrm instance is moved around on a same server and classloader could get included twice.


drush/civicrm.drush.inc
```diff
@@ -620,20 +620,31 @@ function civicrm_drush_cache_clear(&$types) {
...
+  }
 
-  CRM_Core_BAO_WordReplacement::rebuild();
+  if (drush_get_option('sessions', FALSE)) {
+    $params['session'] = 1;
+  }
+
+  $result = civicrm_api3('System', 'flush', $params);
+
+  if ($result['is_error']) {
```
eileenmcnaughton, line 642: you should use a try/ catch block around civicrm_api3 call as it won't return an array if it hits an error

mlutfy:
emilyf reported on IRC that this does not fix the issue for her. Drupal gets bootstrapped fairly early and does the classloader error.

PR #121, opened 2014-02-16
7.x master

ChrisChinchilla:
Fixes to issue - http://issues.civicrm.org/jira/browse/CRM-14072


kurund:
@ChrisChinchilla isn't this duplicate of https://github.com/civicrm/civicrm-drupal/pull/128 ??

ChrisChinchilla:
Err, not sure, I'm not sure why it is, if you think it is, then leave as is, my git knowledge is still a bit shakey.

PR #123, opened 2014-02-20
CRM-14251 views integration:event links - return empty instead of non-li...

eileenmcnaughton:
...nk

text where link is not permissioned


JKingsnorth:
Looks like the D6 version of this PR was pulled in. This is identical so I assume it could also be pulled in?

PR #126, opened 2014-02-27
fix function getOperationPair - not static in parent class

eileenmcnaughton:


eileenmcnaughton:
merge to 4.4 as bug in 4.4

PR #129, opened 2014-03-18
fix civicrm-sql-connect and civicrm-sql-query commands

joelcollinsdc:
There is still a lot of cleanup necessary for drush support to be completely working.  I'm not sure where things stopped working, but I tested with both drush 5 and 6 and neither seemed to work for me.


kurund:
@mlutfy can you please review this.

PR #130, opened 2014-04-07
CRM-14437 CiviEngage walklist fixes

eileenmcnaughton:


davecivicrm:
Note that this PR is against master and the fix will be part of 4.5.

eileenmcnaughton:
that was a mistake - never mind - we have it in our 4.4 now

PR #141, opened 2014-05-03
Crm 13666

ChrisChinchilla:
As discussed here - https://issues.civicrm.org/jira/browse/CRM-13666


colemanw:
Updated PR: https://github.com/civicrm/civicrm-drupal/pull/234

PR #152, opened 2014-07-27
Removed code that was returning values based on views label

ChrisChinchilla:


ChrisChinchilla:
https://issues.civicrm.org/jira/browse/CRM-15025?filter=-1

PR #153, opened 2014-08-04
CRM-15072 - use contact reference for staff responsible group

jmcclelland:
---
- CRM-15072: staff responsible and event contact fields should be contact reference fields not option groups
  https://issues.civicrm.org/jira/browse/CRM-15072

### DIFF
--- a/modules/civicrm_engage/CustomGroupData.xml
+++ b/modules/civicrm_engage/CustomGroupData.xml
@@ -250,8 +250,10 @@
     </CustomField>
     <CustomField>
       <label>Staff Responsible</label>
-      <data_type>String</data_type>
-      <html_type>Select</html_type>
+      <name>staff_responsible</name>
+      <data_type>ContactReference</data_type>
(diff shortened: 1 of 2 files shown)


colemanw:
Seems like this will need an update script.

jmcclelland:
I think making this a "new sites only" change would be reasonable. I think any kind of modification of existing sites would be too difficult since it would be impossible to map option values (which could just have a first name) with contacts in the database.

colemanw:
Yea that makes sense. Since it's just a custom field should we assume this code change will not break existing sites?

jmcclelland:
Yes - definitely. It is only executed when the user manually goes to the configuration screen and selects the option to add the custom data. The option reads:

Civicrm Engage depends on a number of custom data groups and fields and profiles. Check this box to add them. This step is not reversible and will take several minutes to complete."

There is a check to see if this step has been taken, in which case, the following is added: "It appears that the custom data is present."

And... if someone does it anyway, they'll just get errors about how these custom fields have already been created.

PR #166, opened 2014-09-19
Use UFGroup name instead of title in Drupal profile url

samuelsov:
Use profile group name instead of the title whenever we need something that doesn't change depending on the language (url, form data). cf. https://issues.civicrm.org/jira/browse/CRM-15330


davecivicrm:
test this please

eileenmcnaughton:
I'm a bit confused about what needs applying - is it the 3 patches (rebased over master ?) + the one in civicrm-core

eileenmcnaughton:
Based on applying those 3 patches & the drupal one the language switching seems to work based on the url using English either way - is that what you expect?

ie

# /fr/user/59143/edit/Detailed Personal Information

/en/user/59143/edit/Detailed Personal Information

so, the url isn't translated - but on the bright side it works :-)

Somewhat related - I am running a module with code I got off bgm somewhere or other - I'm not quite sure if your patches supercede the need for this
/**
- Implements hook_theme_registry_alter
  */
  function civicrm_languageswitcher_theme_registry_alter(&$theme_registry) {
  // allowing rewrite of theme_links__locale_block from this module as it is not a theme modification
  // should be better to hook links instead but can't find how
  $path = drupal_get_path('module', 'civicrm_languageswitcher');
  $theme_registry['links__locale_block']['theme path'] = $path;
  $theme_registry['links__locale_block']['function'] = 'civicrm_languageswitcher_links__locale_block';
  }

/**
- Rewrite theme_links__locale_block()
- make the language switcher work with civicrm
  */
  function civicrm_languageswitcher_links__locale_block($variables) {
  if (arg(0) == 'civicrm') {
    foreach ($variables['links'] as $lang => $attr) {
      $query = $_REQUEST;
      unset($query['IDS_request_uri']);
      unset($query['IDS_user_agent']);
      $variables['links'][$lang]['query'] = $query;
    }
  }
  return theme('links', $variables);
  }

samuelsov:
You need to apply the 4 commits provided here and the one in https://github.com/civicrm/civicrm-core/pull/4185

You shouldn't have url with title in it like "/en/user/59143/edit/Detailed Personal Information". 
Instead, you should only have url with ufGroup name in it.  ex:  /en/user/59143/edit/Detailed_personal_information_xx
The page title, the links display and the tabs title should be translated in the current language.
Have you cleared all caches ?

The module from bgm is not related to this issue. The module is (i think) used for language switching to make query parameters follow when switching language.
My patch is used to ensure that : 
1. the url user/59143/edit/Detailed_personal_information_xx is the same for every language
2. the page title gets translated
3. the links and tabs gets translated

eileenmcnaughton:
OK - it seems to be working & yes, i think @mlutfy 's code makes contribution pages etc work in the switcher & I think it should be in core

davecivicrm:
This is working for me as well. Eileen / Samuel / Mathieu - is there a pending PR for contribution pages that we should be looking at for core (per Eileen's comment above)?

PR #168, opened 2014-10-01
CRM-15405: Update logic for how to load the conf_path in drupal

joelcollinsdc:
https://issues.civicrm.org/jira/browse/CRM-15405


colemanw:
It looks like this is addressing the same issue as #271 I don't feel qualified to say which approach is better.

PR #169, opened 2014-10-03
CRM-14707 - Add unit-test for Drupal hook invocation

totten:
- [CRM-14707: Cache function existence in api hooks](https://issues.civicrm.org/jira/browse/CRM-14707)


colemanw:
@totten is this ready to be merged?

totten:
Kind of. IIRC, the code ran, but there were tooling/process issues (updating Jenkins, dev docs, etc), and I haven't had time to follow-through on that. Merging it won't break anything and may be useful for a couple folks... but I hesitated on merging because it would be mostly inert/undocumented code. For the moment, I've pushed it to "Future Version".

totten:
This was previously hard-coded to a particular directory structure. It's updated now to use the more flexible `cv php:boot` (like the documented approach to extension testing). Also, I've added an example test for the submodule `civicrm_member_roles`.

totten:
Note: This was a non-functional change in that it doesn't modify any live code but adds a new (currently inactive) test-suite. There's a separate TODO for running the tests on our infra.

PR #178, opened 2014-10-19
CRM-15437 Fix version # in .info files

JKingsnorth:


eileenmcnaughton:
isn't this for 4.5 not master? master must need 4.6

JKingsnorth:
Hi, I've updated this to 4.6, will now re-roll the 7.x branch.

I think it would be better to remove the version number and licence information from the header of all the files though - do we really need all that information at the top of every file?

eileenmcnaughton:
I guess you'd need to ask kurund about that - I don't know the reasons

PR #180, opened 2014-10-24
Metatag fix (tweak)

laryn:
I had to do the ksort separately on its own line or it was just hashing the number 1 for all paths (and returning the same hash, defeating the purpose).


eileenmcnaughton:
Is this still relevant - there have been a few fixes on this line of code... If so maybe open a ticket to link this into as I'm certainly losing track of these fixes...

laryn:
I just took a look at the master branch and someone has already made this change in the interim, so I'll close this one. Thanks.

PR #184, opened 2014-11-05
CRM-15140

kenwest:
Recognise that ksort() returns a boolean rather than an array; prevent PHP generating a notice by replacing !empty() with isset(); and prevent a potential side-effect by not sorting $_GET (lest some code depends on its current order)


kenwest:
Undid the change to isset().

PR #185, opened 2014-11-06
Fix drush civicrm_sql* functionality for drush version >= 6

mollux:
see https://issues.civicrm.org/jira/browse/CRM-15567 for the explanation


monishdeb:
Closing in favor of https://github.com/civicrm/civicrm-drupal/pull/205

PR #188, opened 2014-11-13
CRM-10189 - Drupal Rules Integration - Contact created/updated/deleted...

skellystarman:
... event not firing

This fix was never applied to the Drupal6 stream so applying it now.

---
- CRM-10189: Contact created/updated/deleted event not firing
  https://issues.civicrm.org/jira/browse/CRM-10189

### DIFF
--- a/modules/civicrm_rules/civicrm_rules.module
+++ b/modules/civicrm_rules/civicrm_rules.module
@@ -44,9 +44,16 @@ function civicrm_rules_civicrm_pre($op, $objectName, $objectId, &$objectRef) {
   if (!module_exists('rules') || $op != 'delete') {
     return;
   }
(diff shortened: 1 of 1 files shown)


skellystarman:
QUESTION: I'd like to see this back ported to the 4.3, 4.4 and 4.5 branches also (as they seem to be still maintained, yes?) Is that something I would do and generate separate pull requests for each, effectively duplicating what I have done here for each branch or is that something your merge procedures do anyway? Happy to create the additional PR's.

skellystarman:
Went ahead and created additional PRs for this issue against older branches anyway.

:-)

eileenmcnaughton:
merged into 4.4 & master  - I'm not in a position to make a call on 4.3 & 4.5

PR #191, opened 2014-11-18
CRM-10189 - Drupal Rules Integration - Contact created/updated/deleted ...

skellystarman:
...event not firing

This fix was never applied to the Drupal6 stream so applying it now.

---
- CRM-10189: Contact created/updated/deleted event not firing
  https://issues.civicrm.org/jira/browse/CRM-10189


colemanw:
See #331

PR #201, opened 2014-12-01
CRM-15644 - fix civicrm_group table missing error on login when CiviCRM ...

davejenx:
...DB is separate from Drupal's

---
- CRM-15644: civicrm_group table "missing" error on login when CiviCRM DB is separate from Drupal's
  https://issues.civicrm.org/jira/browse/CRM-15644


eileenmcnaughton:
Monish - do you see any reason why I shouldn't merge this to 4.4? I've looked at the discussion and I see it has been tested against 4.4 & people seem to be in general agreement about it being good

monishdeb:
Eileen - Sure go ahead :)

PR #215, opened 2015-01-26
CRM-15871 also unset q param

eileenmcnaughton:


eileenmcnaughton:
hmm - just thinking about whether this had to go - the addition of the q param had me confused - but I think it was just 'messy' & removing was not required. Will leave this 'hanging around' for a bit.

eileenmcnaughton:
don't think this is needed

PR #219, opened 2015-02-03
CRM-15911: Remove unused api.php

JKingsnorth:


totten:
+1 - I can't see how this file or any of its functions are used.

JKingsnorth:
Great, thanks folks.

PR #225, opened 2015-02-11
CRM-15398 - Support for price sets in views integration.

davecivicrm:
---
- CRM-15398: Enable Drupal Views integration for Price Sets
  https://issues.civicrm.org/jira/browse/CRM-15398


davecivicrm:
ok to test

PR #245, opened 2015-02-27
CRM-16022. Correct "project" field in CiviCRM info files.

xurizaemon:


colemanw:
Looks like the pr can't be merged because it overlaps too much with the previous one.
In future feel free to open just one PR with 2 commits (and one issue in Jira would be fine too).

xurizaemon:
Updated PR, rearranges the order of the entries to be a bit more consistent too.

PR #248, opened 2015-03-02
CRM-16039: fix metatag hook for i18n with prefix-based translation.

mlutfy:


civicrm.module
```diff
@@ -1428,7 +1428,8 @@ function civicrm_modules_uninstalled($modules) {
  * @param array $cid_parts
  */
 function civicrm_metatag_page_cache_cid_parts_alter(&$cid_parts) {
-  if (!empty($cid_parts['path']) && substr($cid_parts['path'], 0, 7) == 'civicrm') {
+  // Avoid substr() because it will not work with Drupal language prefixes.
+  if (!empty($cid_parts['path']) && strpos($cid_parts['path'], 'civicrm') !== FALSE) {
```
colemanw, line 1432: would it be more accurate to say `strpos($cid_parts['path'], 'civicrm') === 0`? That gives the same functionality as the strpos check. (ensuring it is the start of the string)

mlutfy, line 1432: That's the problem with the substr() check: with i18n and prefix-based language detection, the path will start with, say, /en/civicrm/.

This may match "/blog/civicrm-rocks", but imho it's a pretty small consequence (slightly reduced caching)?

colemanw, line 1432: Ok sounds good.

PR #259, opened 2015-03-16
CRM-15030, CRM-16048. Reapply CRM-15030 which was reversed by CRM-16048.

xurizaemon:
Restore fix from CRM-15030.


colemanw:
@xurizaemon those issues are marked for 4.6 are you sure you meant to submit this against master?

xurizaemon:
you're probably right, will PR there instead

PR #261, opened 2015-03-16
CRM-16122. Expose custom field values in human-friendly format.

xurizaemon:
Having to do implode(explode)) feels a bit hackish ... is this the right way to render those here?


colemanw:
I think the "right" way to do this is to use `CRM_Utils_Array::explodePadded`. It takes care of both trimming and exploding.

PR #267, opened 2015-04-07
Adds command to sync users to contacts, useful for thousands of users to...

AronNovak:
... avoid PHP timeout

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -249,6 +249,10 @@ function civicrm_drush_command() {
   $items['civicrm-member-records'] = array(
     'description' => "Run the CiviMember UpdateMembershipRecord cron (civicrm-member-records).",
   );
+  $items['civicrm-sync-users-contacts'] = array(
+    'description' => "Synchronize Users to Contacts: CiviCRM will check each user record for a contact record. A new contact record will be created for each user where one does not already exist.",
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
@mlutfy  - any reason not to merge this?

AronNovak:
Does it make sense to spend time to resolve the conflict? In that case I happily proceed with making it merge-able.

colemanw:
I think this is a good addition. Note that the PR will need to be updated to reflect recent changes in core. The function has been changed from `CRM_Core_BAO_CMSUser::synchronize(FALSE);` to `CRM_Utils_System::synchronizeUsers();`

Note that the new function returns an array of status information, which might be nice to output back to the shell (would require a little parsing - [here's an example](https://github.com/civicrm/civicrm-core/blob/master/CRM/Admin/Form/CMSUser.php#L64-L85) of formatting the output for display).

Aside: the reason you're hitting timeout issues is because the function that performs the syncing is grossly inefficient. But fixing it completely would require the ability to do a sql join between the civi and Drupal databases, which isn't always possible. Ah well.

jitendrapurohit:
cherry-picked this and created a new PR at https://github.com/civicrm/civicrm-drupal/pull/373

monishdeb:
Closed in favour of #373

PR #274, opened 2015-04-29
CRM-16354 - Remove cms wysiwyg from core

colemanw:
https://issues.civicrm.org/jira/browse/CRM-16354


colemanw:
Needs to wait until civicrm/civicrm-core#5687 is merged.

PR #275, opened 2015-04-29
CRM-16309 (drupal user edit profile) fix redirect on delete contact imag...

eileenmcnaughton:
...e byt setting context


eileenmcnaughton:
@kurund bit of a long patch to read but seems OK with & without clean urls

PR #279, opened 2015-05-07
CRM-15539 - Add views support for CiviVolunteer project Id

relldoesphp:
- [CRM-15539: No Support for CiviVolunteer in Views](https://issues.civicrm.org/jira/browse/CRM-15539)


eileenmcnaughton:
I take it that is a civivolunteer table? Really it doesn't belong in core....

Have you considered exposing as an entity instead (ie. use civicrm_entity & add the hook to civivolunteer) 

https://www.drupal.org/project/civicrm_entity & https://www.drupal.org/node/2380721

PR #289, opened 2015-06-18
CRM-16710 : Support for line-items in views integration

Valfr35:
New features for Drupal views integration.
- adding possibility to join contributions to line-items
- adding possibility to join line-items to price and prive-value fields
- adding possibility to join contributions and memberships (two-way) if civiMember and civiContribute are both activated.

(feature previously discussed on CRM-15398)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,5 @@
+# civicrm-drupal
+CiviCRM (Drupal Integration) - adding line-items support
+
+# CRM-16710 
+Support for line-items in views integration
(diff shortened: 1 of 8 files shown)


colemanw:
Thanks for this pull request!
It would be wonderful if you could rebase out those extra 2 "merge" commits since they make our source tree a bit more confusing. Generally we prefer PRs from a single branch containing one or more commits but no unnecessary merges.

Valfr35:
Ok, I rebase (squash) all commits into one.

JKingsnorth:
The readme.md changes will probably need to be come out of the commit? It's looking good though, I was trying to do this just the other day!

Valfr35:
- The readme file is removed
- 1 commit for adding a relationship from price field to price set".

PR #296, opened 2015-07-22
remove reference to patch as in core

eileenmcnaughton:
See https://github.com/civicrm/civicrm-core/pull/6279/files for more

@torrance can you please confirm I'm right about the above PR making this complete


eileenmcnaughton:
Note that with this patch & this PR https://github.com/civicrm/civicrm-buildkit/pull/165 it's possible to load create a d8 site using buildkit

PR #302, opened 2015-07-28
In the absence of memberships, don't assume to remove all roles for which their are synchronisation rules.

torrance:


torrance:
https://issues.civicrm.org/jira/browse/CRM-16932

eileenmcnaughton:
@lolas-freeform @davejenx  can you comment on this in light of your interaction with https://issues.civicrm.org/jira/browse/CRM-14510

davejenx:
I'm not in favour of this change. The established behaviour of civicrm_member_roles for a long time has been that a sync'd role is revoked when a membership is deleted. This makes sense to me. The use case of the module is to grant a role that confers privileges, when the user has a current membership of the relevant type. If they cease to have a current membership, whether by expiry or deletion, they no longer warrant the role. My understanding has always been that sync'd roles should be seen as under the module's control; manually granting or revoking will not persist when the module next syncs.

The use case of wanting to manually assign a role to someone who does not have a membership could be handled by using a different role that confers the same permissions. I think this would be preferable to changing the long-established behaviour of this module, which could be disruptive to existing users.

PR #305, opened 2015-08-14
CRM-6877: Inform Drupal that fields should have no TZ handling.

xurizaemon:
This change appears to resolve TZ issues when displaying CiviCRM Event / Activity dates in Drupal Calendar.

---
- [CRM-6877: Drupal Views Calendar Doesn't Correctly Handle Timezone](https://issues.civicrm.org/jira/browse/CRM-6877)


helenagarcia90:
I've tested the Views Calendar without the patch and it seems to work correctly. Can anyone verify it?

kurund:
@xurizaemon Can you please check @helenagarcia90 comment

PR #312, opened 2015-09-04
drupal_add_js() for drupal 6

Stoob:
coleman said


colemanw:
Unfortunately this won't work. See further details at https://issues.civicrm.org/jira/browse/CRM-17150

PR #315, opened 2015-09-15
CRM-16000 Member Role Sync: only ignore memberships if other of same …

agh1:
…type has lower weight status

---
- CRM-16000: Member Role Sync Uses Inactive Memberships Over Active Memberships
  https://issues.civicrm.org/jira/browse/CRM-16000


agh1:
Closed in favor of https://github.com/civicrm/civicrm-drupal/pull/316

PR #319, opened 2015-10-01
Pretty-print upgrade outputs

eileenmcnaughton:
The upgrade tasks return HTML, so convert to text before printing in the console.

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -693,7 +693,7 @@ function drush_civicrm_upgrade_db() {
     $upgradeHeadless = new CRM_Upgrade_Headless();
     // FIXME Exception handling?
     $result = $upgradeHeadless->run();
-    drush_print("Upgrade outputs: " . "\"" . $result['message'] . "\"");
+    drush_print("Upgrade outputs:\n" . $result['text']);
   }
   else {
     require_once 'CRM/Core/Smarty.php';


eileenmcnaughton:
@mlutfy  happy with this?

colemanw:
This fixes only drush but not wp-cli or... (are their others?)
Maybe the conversion should go farther down into CRM_Upgrade_Headless?

eileenmcnaughton:
hmm - but this is a simple easy win & that is scary code

colemanw:
fair enough.

eileenmcnaughton:
Actually - moving it to this line would probably work https://github.com/civicrm/civicrm-core/blob/master/CRM/Upgrade/Headless.php#L81

But I only know how to test in drupal

colemanw:
I might add a new array key there called "text" and have drush read from that.

eileenmcnaughton:
that sounds pretty easy / safe...

eileenmcnaughton:
I did that here https://github.com/civicrm/civicrm-core/pull/6868 & updated this

eileenmcnaughton:
@colemanw might be worth pinging a WP person to see if they want to adapt this for WP - who is currently active on WP CLI stuff?

eileenmcnaughton:
Core patch https://github.com/civicrm/civicrm-core/pull/6868/commits

PR #321, opened 2015-10-30
CRM-10193 - Allow civicrm-sql-conf and civicrm-sql-cli with empty database

mollux:
[CRM-10193](https://issues.civicrm.org/jira/browse/CRM-10193)

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -202,6 +202,7 @@ function civicrm_drush_command() {
     // explicit callback declaration and non-standard name to avoid collision with "sql-conf"
     'callback' => 'drush_civicrm_sqlconf',
     'description' => 'Print CiviCRM database connection details.',
+    'bootstrap' => DRUSH_BOOTSTRAP_DRUPAL_CONFIGURATION,
   );
   $items['civicrm-sql-connect'] = array(
     // explicit callback declaration and non-standard name to avoid collision with "sql-connect"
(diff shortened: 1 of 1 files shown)


colemanw:
It looks like this changes drush to never bootstrap Civi - aren't there civi drush commands which depend on this?

mollux:
I think this won't break other functionality, as anywhere where the config is really needed, it will be initialized with the CRM_Core_Config::singleton(); method call.

colemanw:
@totten or @eileenmcnaughton any opinion?

eileenmcnaughton:
I don't know that side of things well enough to comment @torrance  or @xurizaemon might - or @mlutfy

jitendrapurohit:
@mollux While doing QA, I found that the above two commands(civicrm-sql-conf and civicrm-sql-cli) works well with the PR. But the upgrade one errors with the message. Eg:-

```
$ drush civicrm-upgrade-db
DB_DataObject Error: No database name / dsn found anywhere
Drush command terminated abnormally due to an unrecoverable error. 
```

Note : With the bootstrap code in place, it works properly.

JoeMurray:
@jlillyreed this seems related to your #340 - could you consider QAing it? Thx

mollux:
@jitendrapurohit There were indeed some commands that need the loaded config, so I made it optional, with the default to load it.This because most of the commands need the loaded config (only the ones using _civicrm_dsn_init don't need it).

jitendrapurohit:
QA'd the changes. All commands seems to be working fine now :-)

PR #324, opened 2015-11-04
Remove core files from drupal branch

eileenmcnaughton:


colemanw:
that's great that they can be removed, I wasn't sure. Why were they there?

eileenmcnaughton:
I'm pretty sure & if they can't I think remove & redress is a better approach in this case anyway

eileenmcnaughton:
Re why they were there - the install instructions used to say "copy this folder to..."

PR #326, opened 2015-11-06
CRM-17517 fix - Drupal integration module references non-existent vie…

monishdeb:
…ws handlers

https://issues.civicrm.org/jira/browse/CRM-17517


colemanw:
FYI @monishdeb this is an extra step (and I don't always do it) but it's nice to give attribution for patches when you commit them by doing e.g.

```
git commit --author "Joe Author <joeauthor@example.com>"
```

cafuego:
@colemanw Oh meh, no worries about that. I should've done a PR if I wanted attribution :-)

monishdeb:
Thanks @colemanw, I will keep that in mind :)

PR #329, opened 2015-11-20
8.x master

andy-walker:
Changed some calls to _drupal_add_html_head() and \Drupal\Component\Utility\SafeMarkup::set(), as they've been removed from the latest 8.x release.


colemanw:
Thanks for the PR! This fixes https://issues.civicrm.org/jira/browse/CRM-17598

PR #338, opened 2015-12-22
CRM-17742 - Added load_generated_data option when installing CiviCRM.

mollux:
[CRM-17742](https://issues.civicrm.org/jira/browse/CRM-17742)


drush/civicrm.drush.inc
```diff
@@ -408,6 +410,9 @@ function _civicrm_install_db($dbuser, $dbpass, $dbhost, $dbname,
   civicrm_source($dsn, $data_file);
   civicrm_source($dsn, $acl_file);
 
+  if ($loadGeneratedData) {
+    civicrm_source($dsn, $sqlPath . DIRECTORY_SEPARATOR . 'civicrm_generated.mysql', TRUE);
```
totten, line 414: This is a bit different from what the web installer does:

https://github.com/totten/civicrm-core/blob/master/install/civicrm.php#L98

In the web-installer, it loads `civicrm_generated.mysql` XOR `civicrm_data.mysql`. In this case, it loads `civicrm_data.mysql` AND `civicrm_generated.mysql`.

PR #342, opened 2016-01-06
Text string unexpected cut at line 56

yurg:
Please add missed string part at line 56, I've closed line properly since it prevents civicrm database update via drush.


mollux:
This was fixed in #343 for 4.7 and will (hopefully) be fixed in #350 for 4.6

PR #347, opened 2016-01-29
patches for master

julialongtin:
I'm fixing civicrm's support for drupal8. the Civicrm.php change allows civicrm to see which contact is logged in, and the second one allows for my recursive invoke hack (which is not yet in civicrm) to function.


colemanw:
I think #348 solves this.

PR #348, opened 2016-02-03
CRM-17609: Corrected retrieval of current user uid.

twomice:
- [CRM-17609: D8 - Synchronize Users to Contacts broken](https://issues.civicrm.org/jira/browse/CRM-17609)


twomice:
CRM-17609

PR #349, opened 2016-02-03
CRM-17966: Removed reference to non-existent Drupal method.

twomice:
- [CRM-17966: Fatal error in Drupal 8 user creation](https://issues.civicrm.org/jira/browse/CRM-17966)


twomice:
CRM-17966

PR #361, opened 2016-03-11
CRM-18222: Allow installation in /modules/civicrm.

mlutfy:
- [CRM-18222: Drupal 8 support: allow installation in \/modules](https://issues.civicrm.org/jira/browse/CRM-18222)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Installation Steps
 
 - Download the latest Drupal 8 release: (from https://www.drupal.org/project/drupal).
 - Install Drupal 8 (see https://api.drupal.org/api/drupal/core!INSTALL.txt/8 for more information).
-- Create a top level `libraries` folder inside Drupal and download CiviCRM master (for Drupal 7) (from http://dist.civicrm.org/by-date/latest/master/) so that CiviCRM resides at `/libraries/civicrm`.
+- Create a top level `libraries` folder inside Drupal and download CiviCRM master (for Drupal 7) (from http://dist.civicrm.org/by-date/latest/master/) so that CiviCRM resides at `/libraries/civicrm`. You can also install in /modules/civicrm (see CRM-18222).
 - Remove the Drupal 7 module folder from within CiviCRM (`libraries/civicrm/drupal`).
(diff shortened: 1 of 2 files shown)


totten:
+1

Ping @torrance, @xurizaemon. My sense is that we should support the old directory structure for now since we know how to ship that, and we haven't found an easy way to make `libraries` work for users. I like this patch in that it seems like we could support both the Old Way and the Drupal Way.

xurizaemon:
:+1: to being able to use both `/libraries/civicrm` and `/modules/civicrm`.

Drupal 8 I believe permits subfolders, so as well as `/profiles` we might see eg `/modules/contrib/civicrm`. 

I feel like CiviCRM shouldn't need to do its own discovery here and we should use [`drupal_get_path`](https://api.drupal.org/api/drupal/core%21includes%21bootstrap.inc/function/drupal_get_path/8) / [`libraries_get_path`](http://www.drupalcontrib.org/api/drupal/contributions!libraries!libraries.module/function/libraries_get_path/8) in Drupal space.

@mlutfy how does the PR below look to you? Untested, will do so when :no_mobile_phones: :smile: 

Not 100% on why Drupal has its own `realpath` but it does so ...

mlutfy:
@xurizaemon Seems good! Tested locally with Drupal in /modules.

mlutfy:
@xurizaemon Did you open a PR for your patch?

tamcy:
Sorry to reply to a closed issue, I am not sure if it is appropriate to do so (this is the first time I give CiviCRM a try). But I am afraid there is an issue with commit 9921231. Specifically, `_civicrm_find_civicrm()` now attempts to call `\Drupal\Core\File\FileSystem::realpath($path)` as if it is a static method, but [the source](https://github.com/drupal/drupal/blob/8.0.5/core/lib/Drupal/Core/File/FileSystem.php#L126) shows that the called method is actually a class/instance method. `FileSystem::realpath` also accesses instance variable (`$this->streamWrapperManager->getViaUri()`). As a result, installation fails with the following error message:

```
Using $this when not in object context in /..../core/lib/Drupal/Core/File/FileSystem.php on line 130
Drupal\Core\File\FileSystem::realpath() /..../modules/civicrm/civicrm.install:135
```

Thanks.

colemanw:
@tamcy thanks. Would you be able to open a pull-request to fix that problem?

PR #363, opened 2016-03-17
CRM-18222. Use Drupal functions for discovery of CiviCRM install location.

xurizaemon:
Refs #361.

---
- [CRM-18222: Drupal 8 support: allow installation in \/modules](https://issues.civicrm.org/jira/browse/CRM-18222)


mlutfy:
Thanks!

PR #372, opened 2016-04-11
Don't look for Wordpress in a Drupal installation

JKingsnorth:
This file is a bit of a train wreck. This is just a small improvement for now.


totten:
@JKingsnorth "This file is a bit of train wreck" ==> agree. Also, in terms of build-process, it's a bit fiddly that we have lookup/copy a CMS-specific variant of of `civicrm.config.php`.

You might find this interesting:
- https://github.com/civicrm/cv/blob/master/src/Bootstrap.php - Basically does the same thing (locating `civicrm.settings.php`) but supports each CMS in one file.
- https://github.com/civicrm/cv - Bootstrap.php repackaged as a CLI tool.

PR #375, opened 2016-04-20
Fix for CRM-18451

laryn:
Only use admin theme if user has access to CiviCRM.

---
- [CRM-18451: Users without access to CiviCRM get admin theme on access denied](https://issues.civicrm.org/jira/browse/CRM-18451)


colemanw:
Reviewed the code changes - seems clear cut. Makes sense to me.

PR #376, opened 2016-04-26
Update civicrm.grant.inc

yurg:
Doesn't work well with boolean value when filtered by "CiviCRM Grants: Grant Report Received": always returns empty results. Changing this to numeric makes it work as needed. Not sure if numeric is the best approach here, but it works.


monishdeb:
screen shot 2016-05-20 at 2 41 29 pm
I was not able to replicate the issue on my local please check the screenshot

yurg:
Wish I remembered where and why it has been made. OK, till next time: changes I've made will be overridden by next civi update so I shall got corresponded complains than.

PR #377, opened 2016-04-26
Remove pass_by_ref & which causes strict error

eileenmcnaughton:
Ref https://github.com/civicrm/civicrm-core/pull/8243


totten:
Closed in favor of #378.

PR #383, opened 2016-05-09
Replace IN to MANY_TO_ONE

yurg:
It seems that many_to_one provides more options at exposed filter operators level; namely it adds "Is all of" option which makes Tags search return only contacts  tagged by ALL tags; useful  for multiply tags search. Not sure about caveats, there must be some.


colemanw:
Sorry, the Drupal 6 branch of this project is not maintained, so this PR has lingered for a while. If you can find someone to review it, great. Please resubmit to the 6.x-master branch.

PR #384, opened 2016-05-25
CRM-17789 - Support PHP 7 - mysqli support in _civicrm_install_db

mollux:
- [CRM-17789: Support PHP 7](https://issues.civicrm.org/jira/browse/CRM-17789)


totten:
I merged this manually and included a follow-up change, 660106de95.

PR #387, opened 2016-06-10
revert #282 to fix ajax uploads on forms

jitendrapurohit:
revert #282 to fix https://issues.civicrm.org/jira/browse/CRM-16513?focusedCommentId=90306&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-90306

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -850,7 +850,7 @@ function civicrm_add_jquery(&$html) {
   return $html;
 }
 
-function civicrm_form_alter(&$form, &$formValues, $formID) {
+function civicrm_form_alter(&$form, $formValues, $formID) {
   switch ($formID) {
     case 'user_admin_permissions':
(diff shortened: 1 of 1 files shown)


JoeMurray:
@samuelsov what about reviewing this PR - seems much simpler in terms of its history.

samuelsov:
It's an exact rollback a the PR-282. So i guess it's a no brainer.
It won't solve the issue CRM-16513 but it should solve the regression mentioned.
I will do a test to see if the PR-282 was creating the problem on the first hand.

samuelsov:
Ok, i have replicated the problem by adding a civicrm profile in the user/register page + a custom image field (added from admin/config/people/accounts/fields) and i confirm that the reverting of #282 resolves this.

This PR has been tested and is ready to be commited.

PR #388, opened 2016-06-17
CRM-17917 - Multiple fixes for fresh Drupal 8 installation

totten:
See detailed descriptions in commit messages.

---
- [CRM-17917: Add support for Drupal 8](https://issues.civicrm.org/jira/browse/CRM-17917)

### DIFF
--- a/src/Controller/CivicrmController.php
+++ b/src/Controller/CivicrmController.php
@@ -29,17 +29,9 @@ class CivicrmController extends ControllerBase {
     );
   }
 
-  public function main($args, $one, $two, $three, $four, $five) {
-    // $one, $two... $five are optional path components. These default to an empty
-    // string. We append any non-empty path components to the $args array before passing
-    // this to invoke().
-    foreach ([$one, $two, $three, $four, $five] as $optional_arg) {
-      if ($optional_arg) {
-        $args[] = $optional_arg;
-      }
-      else {
-        break;
-      }
+  public function main($args, $extra) {
+    if ($extra) {
+      $args = array_merge($args, explode('/', $extra));
     }
 
     // CiviCRM's Invoke.php has hardwired in the expectation that the query parameter 'q' is being used.
--- a/civicrm.install
+++ b/civicrm.install
@@ -126,13 +126,13 @@ function civicrm_requirements($phase) {
 function _civicrm_find_civicrm() {
   if ($path = drupal_get_path('module', 'civicrm')) {
     if (file_exists($path . '/CRM/Core/ClassLoader.php')) {
-      return \Drupal\Core\File\FileSystem::realpath($path);
+      return realpath($path);
     }
   }
 
   if ($path = libraries_get_path('civicrm')) {
     if (file_exists($path . '/CRM/Core/ClassLoader.php')) {
-      return \Drupal\Core\File\FileSystem::realpath($path);
+      return realpath($path);
     }
   }
 
--- a/src/Routing/Routes.php
+++ b/src/Routing/Routes.php
@@ -19,19 +19,16 @@ class Routes {
     // and let each default to empty string.
     foreach ($items as $path => $item) {
       $route = new Route(
-        '/' . $path . '/{one}/{two}/{three}/{four}/{five}',
+        '/' . $path . '/{extra}',
         array(
           '_title' => isset($item['title']) ? $item['title'] : 'CiviCRM',
           '_controller' => 'Drupal\civicrm\Controller\CivicrmController::main',
           'args' => explode('/', $path),
-          'one' => '',
-          'two' => '',
-          'three' => '',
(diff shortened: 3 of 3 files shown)


totten:
To setup a new D8 build, I'd suggest at time of writing:

```
civibuild create d8-master --patch https://github.com/civicrm/civicrm-drupal/pull/388 --patch https://github.com/civicrm/civicrm-core/pull/8592
```

seamuslee001:
Just noting to set up a new D8 Build now you just need 

`civibuild create d8-master --patch https://github.com/civicrm/civicrm-drupal/pull/388`

src/Controller/CivicrmController.php
```diff
@@ -29,17 +29,9 @@ public static function create(ContainerInterface $container) {
     );
   }
 
-  public function main($args, $one, $two, $three, $four, $five) {
```
seamuslee001, line 32: @totten After i reverted this and applied a small patch to the crmNavigationMenu smarty function i got the civicrm menu reliably loading

totten: @seamuslee001 , clarification: was that reverting a specific line or all of 495e085ddde163e1078865a17f60d07127bd7d51?

totten: Also: is the D8 build 8.0, 8.1, or something else? (See `core/CHANGELOG.txt`.) IIRC, I needed the patch in 8.1 to get any routes working...

seamuslee001: @totten yes and i built this off 8.2 using the command above.

seamuslee001: clarification it was reverting your whole commit 495e085.

eileenmcnaughton:
@totten any reason not to merge this? My position on 8.x at the moment is that we should merge a bit more agressively if it is to help get it going

seamuslee001:
@eileenmcnaughton The only issue is that i built off 8.2 using buildkit and had to revert one of Tim's commits to make things work right.

eileenmcnaughton:
one of the commits in this PR? Not a good advertisement for it then...

jackrabbithanna:
@seamuslee001 Can you tell me what "crmNavigationMenu smarty function " you had to patch, and how to get this to work?

jackrabbithanna:
@totten , @seamuslee001  , @eileenmcnaughton 
Tested these changes...
I found that totten's patches to src/Routing/Route.php and src/Controller/CivicrmController.php
are necessary, and work. 
the changes in civicrm.install could be better, I like this PR here for that:
https://github.com/civicrm/civicrm-drupal/pull/391 are what should be

Successfully installed CiviCRM into Drupal 8 without civibuild with the changes in the PR and the one mentioned previously.

Drupal 8.2.5
Libraries 8.x-3.x-dev
Jan 4 daily tarball of 4.7

Fixing this bit is good, and should be committed....
I have no CiviCRM menu up top on Civi pages, but the paths themselves work.....that should be a separate issue/PR

seamuslee001:
Thanks Mark, THe test that i needed was that the navigation menu url (that gets loaded by ajax) wasn't quite working as there was an // in instead of a language so i got a patch into core already which sets it to be en_US if not set otherwise. If you have tested and it works for you I'm happy for this to go through

jackrabbithanna:
Ok I needed the changes in this PR to just get any CiviCRM page to load....
I still don't see a menu....
In the watchdog log , I'm seeing page not found errors having to do with the ajax path for the menu.....
So maybe what we need here is a tweak to totten's commit....

the page not found is for this path: (I've got the site in localhost, in a subdirectory of htdocs, no vhost entry, htdocs/drupal825)
http://localhost/drupal825/civicrm/ajax/menujs/202/en_US/1/NFmjWSIO

What should the path to the ajax menu callback be?

totten:
Thanks for the feedback @jackrabbithanna. 

On my local D7, the path looks very similar -- in HTML, it's `/civicrm/ajax/menujs/203/en_US/1/cIVjCDYN` (which resolves to `http://dmaster.l/civicrm/ajax/menujs/203/en_US/1/cIVjCDYN`).

The route `civicrm/ajax/menujs` has an uncommon property where it actually does take advantage of multiple slashes of subordinate paths. (You can see the route is actually defined as [`civicrm/ajax/menujs`](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/xml/Menu/Admin.xml#L634).)

This suggests that the option `'extra' => '.*',`  isn't having the intended effect (matching the route to all subpaths)?

See also: http://symfony.com/doc/current/routing/slash_in_parameter.html

colemanw:
Merging as most comments suggest this is an improvement on the whole.

jackrabbithanna:
I'll keep pushing on this in a new PR to get the menu going

PR #392, opened 2016-07-19
Fix variable name in 'drush' update code

kenahoo:
Looks like the variable `$tarpath` is mistakenly referred to as `$tarfile` in the `_civicrm_extract_tarfile` function a few places - if so, this function wouldn't have worked properly since early 2013 (which agrees with my experience).

In addition, since `.tgz` and `.tar.gz` are both common suffixes for gzipped tar files, the code now properly handles both variants.

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -370,8 +370,8 @@ function drush_civicrm_install() {
 
 function _civicrm_extract_tarfile($destinationPath, $option = 'tarfile') {
   $tarpath = drush_get_option($option, FALSE);
-  if (drush_shell_exec("gzip -d " . $tarfile)) {
-    $tarpath = substr($tarfile, 0, strlen($tarfile) - 3);
+  if (drush_shell_exec("gzip -d " . $tarpath)) {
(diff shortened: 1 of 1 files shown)


kenahoo:
Ping - this looks like a really easy bug to fix, and has been broken a long time.

kenahoo:
Anyone out there?

totten:
Thanks, @kenahoo.

Ordinarily, we would have pinged you last week about organizing 4.7.11, but August is a bit special. (A lot of folks have been on vacation this month.) Some links that may be useful:
- [(civicrm-dev) 4.7.11 Invitation (Sort Of)](https://lists.civicrm.org/lists/arc/civicrm-dev/2016-08/msg00009.html)
- [release-management: README](https://github.com/civicrm/release-management/blob/master/README.md)
- [civicrm-core: CONTRIBUTING.md](https://github.com/civicrm/civicrm-core/blob/master/.github/CONTRIBUTING.md)

This patch looks pretty conservative, and there are only one or two codepaths which hit it, so I'm inclined to merge.

kenahoo:
Excellent, thanks!  Sorry to appear impatient - my first PR in this project and I wasn't sure what the protocols were.

PR #393, opened 2016-07-31
7 x master

pmosey:
Exposes case type description column to drupal views and changes the section heading

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -386,7 +386,7 @@ function _civicrm_install_db($dbuser, $dbpass, $dbhost, $dbname,
 
   if (function_exists('mysqli_connect')) {
     $dbhostParts = explode(':', $dbhost);
-    $conn = @mysqli_connect($dbhostParts[0], $dbuser, $dbpass, '', isset($dbhostParts[1]) ? $dbhostParts[1] : '');
+    $conn = @mysqli_connect($dbhostParts[0], $dbuser, $dbpass, '', isset($dbhostParts[1]) ? $dbhostParts[1] : NULL);
     $dbOK = ($a = @mysqli_select_db($conn, $dbname)) || ($b = @mysqli_query($conn, "CREATE DATABASE $dbname"));
   }
   elseif (function_exists('mysql_connect')) {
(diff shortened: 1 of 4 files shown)


drush/civicrm.drush.inc
```diff
@@ -383,15 +383,25 @@ function _civicrm_install_db($dbuser, $dbpass, $dbhost, $dbname,
...
+  else {
+    $dbOK = FALSE;
+  }
+  if (!$dbOK) {
     drush_die(dt('CiviCRM database was not found. Failed to create one.'));
   }
 
   // setup database with civicrm structure and data
-  $dsn = "mysqli://{$dbuser}:{$dbpass}@{$dbhost}/{$dbname}?new_link=true";
+  $dsn = "mysql://{$dbuser}:{$dbpass}@{$dbhost}/{$dbname}?new_link=true";
```
seamuslee001, line 404: @totten Tim do you think that where we have mysqli we should priortise that or not? I think that if we have mysqli the DSN should probably switch to mysqli else fall back to mysql?

totten: @seamuslee001, yes, but this is not the right place to handle that.

Look at this way:
- There's the _normal_ case where we connect to MySQL through the PEAR DB abstractions (e.g. `DB::connect()`). If there's a DSN like `mysql://user:pass@host/name`, then we're using PEAR DB abstractions. We want to cleanly switch all of these from `mysql://` to `mysqli://` in one go -- which only requires a 1-line patch to `packages/DB.php`. It's much harder to hunt down all the things which refer to `mysql://` and change them one-by-one.
- There are also the evil cases where someone circumvented PEAR DB and connected directly with a lower-level API. The main place we see this is installation code (like lines 386-390). (_Theory_: Circumventing PEAR DB allows precise error-handling in edge-cases like "the user is valid but the database has not been created yet".) 

If the code constructs a DSN like `mysql://`, stick with `mysql://`, then keep it that way. If the code circumvents PEAR DB (e.g. `mysql_connect()`), then it needs a patch. (An adaptive approach which picks between `mysql_*()` and `mysqli_*()` is nice because we don't have to think as hard when juggling PRs across different repo's, but it's not strongly required. The official policy will be that only one driver is supported.)

PR #394, opened 2016-08-29
Prefer mysqli driver over mysql as more future support. Fall back to …

seamuslee001:
…mysql if cannot get mysqli

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -401,7 +401,8 @@ function _civicrm_install_db($dbuser, $dbpass, $dbhost, $dbname,
   }
 
   // setup database with civicrm structure and data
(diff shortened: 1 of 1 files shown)


seamuslee001:
@totten This would mean we prefer for fresh installs to do mysqli driver but if we can't find mysqli then fall back to mysql. To me it just makes it more stable going forward.

totten:
Not necessary. The on-disk format for `CIVICRM_DSN` will remain unchanged (i.e. you create a configuration file which says `mysql://user:pass@host/name`). When migrating between versions/environments, neither the sysadmin nor the installation script should know the difference between `mysql` and `mysqli`.

What will change is how [`DB::parseDSN()`](https://github.com/civicrm/civicrm-packages/blob/4.7.10/DB.php#L743) interprets the DSN. (To wit: `$dsn = preg_replace('/^mysql:/', 'mysqli:', $dsn)`)

PR #396, opened 2016-09-07
Remove error-suppression arroba which keeps errors out of the log as …

universalhandle:
…well as the off the screen and which just cost me an hour of my life.


colemanw:
Makes sense to me.

PR #405, opened 2016-10-19
CRM-19545 Custom field groups missing from views

urlisse:
Fix for CRM-19545 : "Custom field groups which do not directly extend the base 'Contact' disappear from Drupal view after applying patch CRM-18776"

---
- [CRM-19545: Custom field groups which do not directly extend the base 'Contact' disappear from Drupal view after applying patch CRM-18776](https://issues.civicrm.org/jira/browse/CRM-19545)
- [CRM-18776: Views and non-Contact entities with custom field groups](https://issues.civicrm.org/jira/browse/CRM-18776)

### DIFF
--- a/modules/views/components/civicrm.core.inc
+++ b/modules/views/components/civicrm.core.inc
@@ -2911,7 +2911,7 @@ function _civicrm_core_data(&$data, $enabled) {
 
   while ($dao->fetch()) {
     // call getTree using $dao->id as groupID, $dao->extends as entityType, with possible subtypes in $dao->extends_entity_column_value
-    $extendsContactSubtype = (('Contact' === $dao->extends) && !empty($dao->extends_entity_column_value));
+    $extendsContactSubtype = (in_array($dao->extends, array('Contact', 'Individual', 'Organization', 'Household')) && !empty($dao->extends_entity_column_value));
     $contactSubtypes = !$extendsContactSubtype ? NULL : array_filter(explode(CRM_Core_DAO::VALUE_SEPARATOR, $dao->extends_entity_column_value));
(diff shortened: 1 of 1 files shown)


modules/views/components/civicrm.core.inc
```diff
@@ -2911,7 +2911,7 @@ function _civicrm_core_data(&$data, $enabled) {
 
   while ($dao->fetch()) {
     // call getTree using $dao->id as groupID, $dao->extends as entityType, with possible subtypes in $dao->extends_entity_column_value
-    $extendsContactSubtype = (('Contact' === $dao->extends) && !empty($dao->extends_entity_column_value));
+    $extendsContactSubtype = (in_array($dao->extends, array('Contact', 'Individual', 'Organization', 'Household')) && !empty($dao->extends_entity_column_value));
```
monishdeb, line 2914: You can use `CRM_Contact_BAO_ContactType::basicTypePairs()` instead, which will consider **user-defined** contact types and sub-types too.

monishdeb (changes requested):

modules/views/components/civicrm.core.inc
JKingsnorth: The patch, and this proposed alteration does not fix the issue for custom fields assigned to Relationships. I'll take a look at an alternative solution.

JKingsnorth: I've updated the issue in JIRA with some additional comments.

stesi561:
I think this can be closed. It's very old. Last reference to it on "The follow-up patch @adixon refers to is https://github.com/civicrm/civicrm-drupal/pull/405 , which also will not fix the problem (for the reasons mentioned on that PR)."

and 417 got merged.

PR #406, opened 2016-10-26
CRM-19568 Expose processor ID to views

konadave:
- [CRM-19568: Expose recurring contribution processor ID to views](https://issues.civicrm.org/jira/browse/CRM-19568)

### DIFF
--- a/modules/views/components/civicrm.contribute.inc
+++ b/modules/views/components/civicrm.contribute.inc
@@ -943,6 +943,26 @@ function _civicrm_contribute_data(&$data, $enabled) {
     'title' => 'Recurring Contribution End Date',
       'name' => 'end_date',
     ));
+  //Prcocessor ID
+  $data['civicrm_contribution_recur']['processor_id'] = array(
+    'title' => t('Processor ID'),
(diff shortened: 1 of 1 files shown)


monishdeb:
@konadave will it be `$data['civicrm_contribution_recur']['payment_processor_id']` instead ?

konadave:
@monishdeb No.

`processor_id` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL COMMENT 'Possibly needed to store a unique identifier for this recurring payment order - if this is available from the processor??'

monishdeb:
Hmm I think you are right, but not sure about the title because apparently 'Processor ID' mean  the payment processor's ID used for recurring payment instead of unique identifier.

konadave:
This came about as a request from a client...

"How much time would it take to make the Processor ID field from civicrm_contribution_recur available to Views? This field corresponds to Auth.net's ARB's subscription ID, their unique identifier for recurring payment subscriptions."

The processor ID assigned by the payment processor is a unique identifier.

monishdeb:
I see. Tested on my local and it works fine for me. Merging ..

PR #407, opened 2016-10-31
Added support for dropping the CiviCRM database with drush.

mollux:


eileenmcnaughton:
@bgm @jmcclelland  have any thoughts about this? Does the fact that it would also drop CMS tables if in the same DB cause major concerns? Or do we expect drush users to user with care

jmcclelland:
I think I don't understand the purpose of the command. Why would we need civicrm-sql-drop if drush already offers sql-drop which does the same thing? I'm probably missing something.

On the other hand, if civicrm-sql-drop only dropped tables prefixed with civicrm_ it would be quiet useful to me (since powerbase keeps all drupal and civicrm tables in the same database).

mlutfy:
I understand that this mimics other drush commands, such as `civicrm-sql-cli`, which will automatically connect to the CiviCRM database (if separated from the Drupal database). The `sql-drop` command won't work as-is currently, if your CiviCRM db is separate.

However, I agree with Eileen that it could easily be misinterpreted as "drop all civicrm tables only".

Instead of `drush_sql_drop();` I would recommend something (via CRM_Core_DAO) that would do the equivalent of:

`echo 'show tables' | drush sqlc | grep civicrm | grep -v Tables | gawk -v NOFOREIGNCHECK='SET FOREIGN_KEY_CHECKS=0;' 'BEGIN {print NOFOREIGNCHECK}{print "drop table " $1 ";"}'  | drush sqlc`

and also `drop view`.. which would have the added benefit of working for people who have the drupal/civicrm tables in the same database, as Jamie mentions.

mlutfy:
I understand that this mimics other drush commands, such as `civicrm-sql-cli`, which will automatically connect to the CiviCRM database (if separated from the Drupal database). The `sql-drop` command won't work as-is currently, if your CiviCRM db is separate.

However, I agree with Eileen that it could easily be misinterpreted as "drop all civicrm tables only".

Instead of `drush_sql_drop();` I would recommend something (via CRM_Core_DAO) that would do the equivalent of:

`echo 'show tables' | drush sqlc | grep civicrm | grep -v Tables | gawk -v NOFOREIGNCHECK='SET FOREIGN_KEY_CHECKS=0;' 'BEGIN {print NOFOREIGNCHECK}{print "drop table " $1 ";"}'  | drush sqlc`

(and also `drop view`).

stesi561:
This is very old. And hasn't had any attention. The last comment suggested a different approach would be be better/safer.

I think this can be closed

PR #408, opened 2016-10-31
Added support for dropping the CiviCRM database with drush.

mollux:


totten:
FWIW, `setup.sh` has a similar step (e.g. part of resetting the DB when rebuilding with new schema). The step originally dropped the SQL database, but there was pushback from folks who use single-db deployments (e.g. CMS+Civi in same DB), and so now the logic is more like:

https://github.com/civicrm/civicrm-core/blob/4.7.12/bin/setup.sh#L190

totten:
Two comments -- one substantive, one administrative.

Substantively, it is really useful to be able to drop the Civi schema, so the PR called out a good idea, and we incorporated that general idea as part of [December's work on the CiviCRM installer](https://civicrm.org/blog/totten/developers-revising-the-civicrm-installer-library-cli-wordpress) -- specifically, there's a new subcommand `cv core:uninstall` which will remove all `civicrm_*` tables and and the settings file. Depending on your exact use-case, this might solve the same problem as #408 -- or it might not. They're similar at a high-level (both generally remove Civi tables) but differ in other ways (eg when used on a single-DB deployment, dropping `civicrm_*` tables is more conservative than dropping the whole DB; eg drush has more nuanced support for Drupal bootstrap, and cv has support for more CMSs).

Administratively, the Civi-D8 code is being migrated to a new repository with a slightly different branch naming convention (ie `civicrm-drupal@8.x-master` becomes `civicrm-drupal-8@master`). 
This change facilitates distribution via `composer`.  The new repo is up now, but it means we need to close any existing PRs on `civicrm-drupal@8.x-master` .

If you still think a patch like this is needed, feel free to resubmit against https://github.com/civicrm/civicrm-drupal-8/ . Or if you think that `core:uninstall` has captured enough of the spirit, then I hope it works.

PR #409, opened 2016-11-04
CRM-19604: civicrm-ext-list: avoid default API limit of 25 results.

mlutfy:
* [CRM-19604: civicrm-ext-list: avoid default API limit of 25 results](https://issues.civicrm.org/jira/browse/CRM-19604)

---

 * [CRM-19404: Recommend PHP 5.3.23](https://issues.civicrm.org/jira/browse/CRM-19404)

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -576,7 +576,7 @@ function civicrm_drush_help($section) {
 function drush_civicrm_ext_list() {
   civicrm_initialize();
   try{
-    $result = civicrm_api('extension', 'get', array('version' => 3));
+    $result = civicrm_api('extension', 'get', array('version' => 3, 'option.limit' => 0));
     $rows = array(array(dt('App name'), dt('Status')));
     foreach ($result['values'] as $k => $extension_data) {
       $rows[] = array(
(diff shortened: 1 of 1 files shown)


drush/civicrm.drush.inc
```diff
@@ -576,7 +576,7 @@ function civicrm_drush_help($section) {
 function drush_civicrm_ext_list() {
   civicrm_initialize();
   try{
-    $result = civicrm_api('extension', 'get', array('version' => 3));
+    $result = civicrm_api('extension', 'get', array('version' => 3, 'option.limit' => 0));
```
monishdeb: what if we use `civicrm_api3` instead as patch is against 7.x-master

seamuslee001: @monishdeb I think that makes sense to me and isn't the format 

``` php
array( 'version' => 3, options => array( 'limit' => 0 ))
```

mlutfy:
Updated my PR as per feedback. Thanks for reviewing :)

mlutfy:
hum the CRM reference in my commit is wrong. Replacing with PR #411.

PR #414, opened 2016-11-25
Performance improvement during module enable/disable.

xurizaemon:
CiviCRM does not need to trigger CRM_Core_Invoke::rebuildMenuAndCaches(TRUE) during Drupal module changes unless one of the changed modules is related to CiviCRM.

We check the changed modules for (certain) CiviCRM hooks, and only if we find one do we do the rebuild.

CRM-13796

---

 * [CRM-13796: Don't CRM_Core_Invoke::rebuildMenuAndCaches\(TRUE\) on every Drupal module enable \/ disable](https://issues.civicrm.org/jira/browse/CRM-13796)

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -1249,37 +1249,80 @@ function civicrm_reviews() {
 }
 
 /**
+ * If any of the modules with changed status implement CiviCRM hooks, rebuild
+ * the CiviCRM menu and caches. Otherwise let it pass.
+ *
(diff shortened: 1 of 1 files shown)


totten:
Added review notes in JIRA: https://issues.civicrm.org/jira/browse/CRM-13796

colemanw:
Closing based on review notes.

PR #416, opened 2016-12-05
mysql->mysqli

johanv:
I think this fixes CRM-17874.

---

 * [CRM-17874: drush civicrm-install creates deprecated mysql:\/\/ configuration ](https://issues.civicrm.org/jira/browse/CRM-17874)

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -401,7 +401,7 @@ function _civicrm_install_db($dbuser, $dbpass, $dbhost, $dbname,
   }
 
   // setup database with civicrm structure and data
-  $dsn = "mysql://{$dbuser}:{$dbpass}@{$dbhost}/{$dbname}?new_link=true";
+  $dsn = "mysqli://{$dbuser}:{$dbpass}@{$dbhost}/{$dbname}?new_link=true";
   drush_log(dt("Loading CiviCRM database structure .."));
   civicrm_source($dsn, $sqlPath . '/civicrm.mysql');
   drush_log(dt("Loading CiviCRM database with required data .."));


totten:
Thanks for finding and reporting. :)

FWIW, the DSN is supposed to be interpreted dynamically -- e.g. one always writes `mysql://` in `CIVICRM_DSN`, but the runtime determines whether to use `php-mysql` or `php-mysqli`.

@johanv , could you try https://github.com/civicrm/civicrm-core/pull/9472 ?

johanv:
I tried the patch in civicrm/civicrm-core#9472, but that did not fix it for me:

```
$ drush --include=sites/all/modules/civicrm/drupal/drush cvi --dbuser=dev --dbpass=*** --dbhost=db --dbname=civi --destination=./sites/all/modules --site_url=localhost --lang=nl_NL --langtarfile=$L10N/civicrm-4.7.13-l10n.tar.gz
WD php: Error: Call to undefined function drush_sql_get_class() in drush_civicrm_get_db_spec() (line 271 of /var/www/html/sites/all/modules/civicrm/drupal/drush/civicrm.drush.inc).                                     [error]
Error: Call to undefined function drush_sql_get_class() in drush_civicrm_get_db_spec() (line 271 of /var/www/html/sites/all/modules/civicrm/drupal/drush/civicrm.drush.inc).
Drush command terminated abnormally due to an unrecoverable error. 
```

johanv:
Ok, my bad. I was using a wrong version of drush (drush7 instead of drush8). With drush 8, civicrm/civicrm-core#9472 works.

seamuslee001:
@totten @yashodha I believe this can be closed now as core 9472 has been merged

totten:
Thanks, @joahnv and @seamuslee001 .

PR #418, opened 2016-12-27
CRM-17789 mysqli support in 4.6 to aid support for php7

seamuslee001:
* [CRM-17789: Support PHP 7](https://issues.civicrm.org/jira/browse/CRM-17789)


jitendrapurohit:
This part works fine and creates/selects a db in php7 correctly.

jitendrapurohit (approved):

PR #420, opened 2017-01-07
CiviCRM Menu -- Add PathProcessor to allow parameters to routes to include slashes

jackrabbithanna:
So Drupal 8 doesn't allow route parameters with slashes without some trickery. 

You have to add a PathProcessor, and convert slashes to another character, in this case ':'

When we identify other paths that have parameters with slashes, we should add checks to the path processor and do a conversion....

### DIFF
--- a/src/PathProcessor/CivicrmPathProcessor.php
+++ b/src/PathProcessor/CivicrmPathProcessor.php
@@ -12,7 +12,6 @@ class CivicrmPathProcessor implements InboundPathProcessorInterface {
   public function processInbound($path, Request $request) {
     // if the path is a civicrm path   
     if (strpos($path, '/civicrm/') === 0) {
-      
       // initialize civicrm
       $civicrm = new Civicrm();
       // fetch civicrm menu items
@@ -40,8 +39,6 @@ class CivicrmPathProcessor implements InboundPathProcessorInterface {
         else {
           return $longest;
         }
-       
-   
       }
     }
     return $path;
--- a/src/PathProcessor/CivicrmPathProcessor.php
+++ b/src/PathProcessor/CivicrmPathProcessor.php
@@ -30,10 +30,13 @@ class CivicrmPathProcessor implements InboundPathProcessorInterface {
       }
       if (!empty($longest)) {
         // parse url component parameters from path
-        $params = str_replace($longest . '/', '', $path);
+        $params = str_replace($longest, '', $path);
         // replace slashes with colons and the controller will piece it back together
         if(strlen($params)) {
           $params = str_replace('/', ':', $params);
+          if (substr($params, 0, 1) == ':') {
+           $params = substr($params, 1);
+          }
           return "$longest/$params";
         }
         else {
--- a/src/PathProcessor/CivicrmPathProcessor.php
+++ b/src/PathProcessor/CivicrmPathProcessor.php
@@ -5,15 +5,29 @@ namespace Drupal\civicrm\PathProcessor;
 use Drupal\Core\PathProcessor\InboundPathProcessorInterface;
 use Symfony\Component\HttpFoundation\Request;
 
+use Drupal\civicrm\Civicrm;
+
 class CivicrmPathProcessor implements InboundPathProcessorInterface {
 
   public function processInbound($path, Request $request) {
-    if (strpos($path, '/civicrm/ajax/menujs') === 0) {
-      $params = preg_replace('|^\/civicrm\/ajax\/menujs\/|', '', $path);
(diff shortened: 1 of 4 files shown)


colemanw:
In this case, there wasn't any particular reason for using slashes instead of the traditional `?arg1=` so if it causes this much grief for D8 I'd vote to just change the path in CiviCRM core to something less weird.
On the server-side, all the args are ignored anyway, they are only there to prevent browser caching.

colemanw:
Here is an alternate solution: https://github.com/civicrm/civicrm-core/pull/9650

jackrabbithanna:
From my perspective its six one way and half a dozen the other, just want to make a Drupal 8 integration possible work. 

The only reason to use this PR and include a PathProcessor service, is if there are other paths which will need this handling. 
ATM any civicrm path that take url components instead of ?arg style parameters will result in a 404 in Drupal

If there are other situations where it is not so convenient to change to using ?arg style parameters, something like this will need to be anyway....

colemanw:
One other that comes to mind is `civicrm/admin/options/{option_group}`.

jackrabbithanna:
I think it will be possible to automate CivicrmPathProcessor::processInbound() 

For every path we can can check it to see if its a registered Civi route, but getting the menu items array with \CRM_Core_Menu::items()

If it is a civi path, then we look to see if there are additional path components, and if so, process them...

The controller mods will work already with this....

So maybe we can allow these url component params with any civi path, in one swoop...

This is what I had in mind with https://issues.civicrm.org/jira/browse/CRM-19843

colemanw:
@jackrabbithanna good plan! Go for it. :+1:

jackrabbithanna:
all right, just made another commit, its working for the menu, and that should do it!

src/PathProcessor/CivicrmPathProcessor.php
```diff
@@ -0,0 +1,33 @@
...
+  public function processInbound($path, Request $request) {
+    // if the path is a civicrm path   
+    if (strpos($path, '/civicrm/') === 0) {
+      // initialize civicrm
+      $civicrm = new Civicrm();
+      // fetch civicrm menu items
+      $items = \CRM_Core_Menu::items();
+      foreach (array_keys($items) as $item) {
+        // if he current path is a civicrm path
+        if(('/' . $item  . '/') == $path) {
```
colemanw: Maybe I'm reading it wrong but substituting values for variables this statement looks like `if ('/civicrm/ajax/menujs/1/en_US/1/foo' == '/civicrm/ajax/menujs/')` which would always evaluate to `false`.

jackrabbithanna:
ok, got an update commit....

Turns out you have to clear drupal caches every change you make to this file

This works for pages and the menu...

But submitting forms is doing somthing strange...

So if you go to the basic contact search:
and click "Find Contact", the results come up, but the path in the browser is:

Not sure if this is a problem in the Drupal code, or a different problem with drupal code, or something else entirely...

civicrm/contact/search//civicrm/contact/search?_qf_Basic_display=true&qfKey=9ecd43b2ffe43cd2c4eff2d83ee153aa_8904

jackrabbithanna:
ok think I got it now, menu works, pages work, forms work....

colemanw:
Cool. And did you try it with e.g. `/civicrm/admin/options/activity_type?reset=1`?

jackrabbithanna:
yes that path works, and since I got that last commit I've been using the site, and all the ajax everywhere seems to be working great. In fact in Drupal 8 it seems quite a bit faster to me than Drupal 7....Its nice

jackrabbithanna:
I'm using the yesterday's 4.7 master tarball...

The API explorer gives a javascript error on page load, and in the firebug console i see:
TypeError: _ is undefined
fieldTpl = _.template($('#api-param-tpl').html()),

What do you think causes that?

colemanw:
@jackrabbithanna this is a bug in D8 only?
_ should be a reference to the `CRM._` variable. Try typing that in your console (CRM dot underscore) and see what you get.

jackrabbithanna:
There are javascript related errors on several pages.  Both the contribution page and event edit pages complain about missing Backbone...
Several other pages have jQuery related errors...
I don't think that these things are related to the routing problem in this PR

jackrabbithanna:
and its not that the backbone script isn't loading, it is...

Civicrm adds its js into the head (backbone 1.0.) , and drupal is adding a different version before the end of the closing body tag (1.2.3)

Will probably make a separate issue in JIRA for javascript stuff, and we may have to do some juggling to make sure civicrm can make use of the version it needs....

PR #423, opened 2017-02-01
CRM 19835 Installing into D8, DB requirements fail using non-standard…

VasanthaKaje:
Removed the port number concatenation, as mysqli_connect api specs are different

[https://github.com/civicrm/civicrm-core/pull/9760](https://github.com/civicrm/civicrm-core/pull/9760) is the associated fix to civicrm-core

### DIFF
--- a/civicrm.install
+++ b/civicrm.install
@@ -85,7 +85,6 @@ function civicrm_requirements($phase) {
 
   // Grab db connection info
   $db_config = _civicrm_get_db_config()['info'];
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
@totten, I think this should get merged in

jackrabbithanna:
@colemanw  what do you think?

totten:
Merging. Supported by @jackrabbithanna and @seamuslee001. Only affects 8.x. Related comments in https://github.com/civicrm/civicrm-core/pull/9910.

PR #426, opened 2017-02-06
CRM-19542 Unable to theme language switcher block

kainuk:
* [CRM-19542: Unable to theme language switcher block](https://issues.civicrm.org/jira/browse/CRM-19542)

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -290,13 +290,12 @@ function civicrm_metatag_metatags_view_alter(&$output, $instance) {
  *
  * @return array
  */
-function civicrm_links__locale_block($variables) {
+function civicrm_preprocess_links__locale_block(&$variables) {
   if (arg(0) == 'civicrm') {
     foreach ($variables['links'] as $lang => $attr) {
       $variables['links'][$lang]['query'] = _civicrm_get_url_parameters();
     }
   }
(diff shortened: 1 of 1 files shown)


colemanw:
@mlutfy do you use this block? Could you please test this patch to ensure it doesn't cause any regression?

jackrabbithanna:
well that patch does make sense

if you can get the same behavior with a preprocess, why have your own theme function, and override the theme registry to use it?

I think this patch may be the way to go, but we need to figure out a way to test if it breaks themes....

@kainuk what theme do you use?

jackrabbithanna:
actually it shouldn't matter what theme. 

it just needs to be tested in a multilingual drupal + multilingual civicrm site, with the Drupal language switcher block placed on civicrm pages

mlutfy:
I tested this on some sites, but it didn't work:  clicking on the language block (from a page in French) would give me an URL such as: `https://demo.symbiotic.coop/fr/civicrm/contribute/transact?q=en/civicrm/contribute/transact&reset=1&id=1`

@kainuk Had this been tested on a multi-lingual site? (with locale+i18n)

stesi561:
Last comment suggests this isn't working. Given that this hasn't been touched for a while now perhaps we can close.

PR #428, opened 2017-02-18
Drupal 8, User View page not showing profiles CRM-20051

jackrabbithanna:
* [CRM-20051: Drupal 8 user view page, Profile information not displaying](https://issues.civicrm.org/jira/browse/CRM-20051)


jackrabbithanna:
@totten , this was a quick an easy fix, after this patch the profile shows on the user view page

ok to merge

jackrabbithanna:
@colemanw what do you think?

PR #429, opened 2017-02-27
CRM-20186 - Expose legal name in views

yashodha:
* [CRM-20186: Expose legal name in views](https://issues.civicrm.org/jira/browse/CRM-20186)


eileenmcnaughton:
This looks fine to me. I admit to not having tested it but I feel that just adding an extra field to the exposed array of fields is low risk and I accept Yashi's testing on it

PR #432, opened 2017-02-27
CRM-20186 - Expose legal name in views

yashodha:
* [CRM-20186: Expose legal name in views](https://issues.civicrm.org/jira/browse/CRM-20186)


stevekessler:
I have tested this and it working well.

PR #433, opened 2017-03-07
CRM-20205 - Expose current employer id to views

yashodha:
* [CRM-20205: Expose current employer id to views](https://issues.civicrm.org/jira/browse/CRM-20205)


stevekessler:
This is tested and works very well.

PR #436, opened 2017-03-17
CRM-20298: Custom fields of type Money should use the same filter operators as Integer fields

universalhandle:
* [CRM-20298: Drupal Views: Custom fields of type Money should use the same filter operators as Integer fields](https://issues.civicrm.org/jira/browse/CRM-20298)

### DIFF
--- a/modules/views/civicrm.views.inc
+++ b/modules/views/civicrm.views.inc
@@ -5305,6 +5305,7 @@ function civicrm_get_filter_handler($type) {
 
     case "Float":
     case "Int":
+    case "Money":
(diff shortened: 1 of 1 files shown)


universalhandle:
I flagged this as a WIP (work in progress) because I was seeing some funny behavior on a client's site after applying the patch. I eventually tracked it down to a problem in custom code, so I believe this is ready for review/merging.

jackrabbithanna:
I recommend merging this

jackrabbithanna:
Merging #436

PR #437, opened 2017-03-22
CRM-20319 - Expose relationship permission to views

yashodha:
* [CRM-20319: Expose relationship permission to views](https://issues.civicrm.org/jira/browse/CRM-20319)


stevekessler:
This is working great for me.

yashodha:
merging as per confirmation from @stevekessler

PR #438, opened 2017-03-22
CRM-20319 - Expose relationship permission to views

yashodha:
* [CRM-20319: Expose relationship permission to views](https://issues.civicrm.org/jira/browse/CRM-20319)


stevekessler:
I have tested this and it works well.  Have the is_permissioned value has been very helpful for us. We have used the Conditional Fields module to show information if this is true.

PR #415, opened 2016-11-29
CRM-19740: Regression in phone number Drupal views integration

Vgratioulet:
Sorry I had two missing files in previous commit to solve https://issues.civicrm.org/jira/browse/CRM-19274

---

 * [    CRM-19759 handles 'require this relationship' for location-related element relationships](https://issues.civicrm.org/jira/browse/CRM-19759)
 * [CRM-19760: add several relationships of the same location-related element](https://issues.civicrm.org/jira/browse/CRM-19760)
 * [CRM-19740: Regression in phone number Drupal views integration](https://issues.civicrm.org/jira/browse/CRM-19740)

PR #439, opened 2017-03-24
CRM-20203: Gmap Views broken (don't find geo_code)

jitendrapurohit:
Not sure if it should be done in core, but it fixes CRM-20203. Due to changes in https://github.com/civicrm/civicrm-drupal/pull/415/files#diff-eb1aa2cc8ada7a16fe8b497a244d9cb7R83, a new join is added for each table field which breaks the alias(civicrm_address_geo_code_2) we select for IGVM Maps. With additional address join, alias is created as civicrm_address2_geo_code_2.

As this was in a working state before, did a PR to fix it in upper versions too.

---

 * [CRM-20203: Gmap Views broken \(don't find geo_code\)](https://issues.civicrm.org/jira/browse/CRM-20203)

### DIFF
--- a/modules/views/civicrm/civicrm_handler_field_address.inc
+++ b/modules/views/civicrm/civicrm_handler_field_address.inc
@@ -80,7 +80,8 @@ class civicrm_handler_field_address extends civicrm_handler_field_location {
         exit;
       }
       $join = $this->get_join();
-      if (empty($this->relationship)) { //https://www.drupal.org/node/2617032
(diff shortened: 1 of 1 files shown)


jitendrapurohit:
Closing as per the comment on https://issues.civicrm.org/jira/browse/CRM-20203?focusedCommentId=102011&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-102011

PR #440, opened 2017-04-21
CRM-20462 - Add filter for current employer in views

yashodha:
* [CRM-20462: Add filter for current employer in views](https://issues.civicrm.org/jira/browse/CRM-20462)


eileenmcnaughton:
@jitendrapurohit perhaps you could review this one & the other one against 7.x-master if you get a chance - and suggest one that @yashodha might review of yours in return

stevekessler:
I have tested it and it works well.

eileenmcnaughton:
Good enough for me!

PR #441, opened 2017-04-21
CRM-20462 - Add filter for current employer in views

yashodha:
* [CRM-20462: Add filter for current employer in views](https://issues.civicrm.org/jira/browse/CRM-20462)


stevekessler:
I have tested this and it works well.

PR #444, opened 2017-04-29
CRM-20502 - Do not get custom group tree into cache if we are upgrading

seamuslee001:
@eileenmcnaughton @totten @karing try this

---

 * [CRM-20502: Drush upgrade crashes on loading "is_public" field](https://issues.civicrm.org/jira/browse/CRM-20502)

### DIFF
--- a/modules/views/civicrm.views.inc
+++ b/modules/views/civicrm.views.inc
@@ -304,6 +304,9 @@ function civicrm_views_add_date_arguments(&$data, $value) {
  *   Array with the new custom field appended
  */
 function civicrm_views_custom_data_cache(&$data, $entity_type, $group_id, $sub_type) {
+  if (!CRM_Core_Config::isUpgradeMode()) {
+    return;
+  }
   // From http://forum.civicrm.org/index.php/topic,17658.msg73901.html#msg73901, CRM-7860.
   $tree = CRM_Core_BAO_CustomGroup::getTree($entity_type, CRM_Core_DAO::$_nullObject, NULL, $group_id, $sub_type, NULL);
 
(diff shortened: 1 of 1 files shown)


modules/views/civicrm.views.inc
```diff
@@ -304,6 +304,9 @@ function civicrm_views_add_date_arguments(&$data, $value) {
  *   Array with the new custom field appended
  */
 function civicrm_views_custom_data_cache(&$data, $entity_type, $group_id, $sub_type) {
+  if (!CRM_Core_Config::isUpgradeMode()) {
```
seamuslee001: @totten should it actually be if (CRM_Core... not the negating?

eileenmcnaughton: I think you are right - this 'works' because it is ALWAYS returning!

KarinG: I commented on Mattermost - but will confirm here too that this indeed does work! :-)

seamuslee001: @KarinG are you able to test changing the if to be if (CRM_Core_Config::isUpgradeMode())

seamuslee001: @KarinG @eileenmcnaughton i have just changed it to remove the ! as i tested it and didn't seem to make a difference and i think is the actual correct thing we should be testing

eileenmcnaughton:
This makes sense but, as discussed can you try to get some more people to test.

eileenmcnaughton:
NB I mean test on the rc with this merged - not the PR

modules/views/civicrm.views.inc
KarinG: @seamuslee001 so if it doesn't seem to make a difference whether ! or not - then how can we get to the rest of this function if we're not upgrading?

seamuslee001: so i tested in upgrade and it made no difference i think that removing the ! will make a difference in non upgrade times

totten: Yes, I think you wound up with the correction formulation (short-circuit the request in upgrade mode).

PR #447, opened 2017-05-18
CRM-20541 - Use drupal_static() instead of static

totten:
The impetus behind CRM-20541 seems to be a testing scenario where the
setup/teardown process resets Civi (`Civi::reset()`) without resetting
Drupal (`drupal_static_reset()`).  IMHO, it's better to keep the systems
aligned by either (a) resetting both or (b) resetting neither.

For the situation where you reset both, the state within
`civicrm_initialize()` needs some way to reset.  This function executes
within a Drupal context (before Civi has booted), so it should obey the
reset conventions for Drupal -- i.e.  use `drupal_static()`.

Ping @eileenmcnaughton

---

 * [CRM-20541: Edge case where DB connection is not available](https://issues.civicrm.org/jira/browse/CRM-20541)

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -156,8 +156,8 @@ function civicrm_initialize() {
   }
   _civicrm_registerClassLoader();
 
-  static $initialized = FALSE;
-  static $failure = FALSE;
+  $initialized = &drupal_static('civicrm_initialize', FALSE);
+  $failure = &drupal_static('civicrm_initialize_failure', FALSE);
 
   if ($failure) {
     return FALSE;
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
I think this seems like real good idea.  If you are using CiviCRM Entity, its a heavy user of civicrm_initialize() and things could be happening during cron, in drush, all kinds of processing before and after page loads. Who knows what never before seen conditions could crop up. Better that is in the management of the Drupal static system.

jackrabbithanna:
I've patched my local dev instance where I'm working on CiviCRM Entity stuff, and I'll watch and see if anything seems to go amiss.  So far works fine (4.6 though), but shouldn't matter really in this case I don't think.

eileenmcnaughton:
I'm going to merge this - not yet sure if it solves all the test issues I'm having, but I'm pretty sure this is more correct and I don't believe there is any risk with it.

PR #455, opened 2017-06-20
WIP: CRM-20751: Support Drupal aliases for event links in Views

twomice:
Darn, need to fix civilint errors.

---

 * [CRM-20751: Support Drupal aliases for event links in Views](https://issues.civicrm.org/jira/browse/CRM-20751)

### DIFF
--- a/modules/views/civicrm.views.inc
+++ b/modules/views/civicrm.views.inc
@@ -142,9 +142,8 @@ function civicrm_views_data_alter(&$data) {
  * @return String path to CiviCRM
  */
 function civicrm_views_href($text, $path, $query) {
-  civicrm_initialize();
-  require_once 'CRM/Utils/System.php';
-  return CRM_Utils_System::href($text, $path, $query);
+  $url = civicrm_views_url($path, $query);
+  return "<a href=\"$url\">$text</a>";
 }
 
 /**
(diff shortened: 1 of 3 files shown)


twomice:
Hi @colemanw I'd like to get your feedback on this please. BTW, there are civilint errors that were already there before; is it preferable for me to clean them up in this PR?

colemanw:
@twomice I'm curious to know why not patch the `civicrm_views_href` and `civicrm_views_url` functions to check drupal path aliases - it's a nice central place to put in this feature.

colemanw:
@twomice for that matter, why not have those functions pass the job off to Drupal and skip the `civicrm_initialize` and `CRM_Utils_` stuff.

twomice:
That would expand the scope to "all of CiviCRM Views" instead of just "Events in Views", but yeah, I'm cool with that. Will give it a try.

colemanw:
Seems like a comfortable middle-ground between "just events in views" 
and "all of CiviCRM core"

twomice:
Jenkins test this please.

twomice:
@colemanw Right, that makes sense. So here it is. This looks right and works for me.

colemanw:
I like it.

PR #456, opened 2017-06-21
CRM-20751 - Handle array query data in civicrm_views_url

colemanw:
* [CRM-20751: Support Drupal aliases for event links in Views](https://issues.civicrm.org/jira/browse/CRM-20751)


colemanw:
@twomice I couldn't find an example of this scenario, but better safe than sorry.

twomice:
Nice, Thanks!

PR #457, opened 2017-06-25
CRM-19976: Fix for Drush: cannot disable civicrm debug

kainuk:
* [CRM-19976: Drush: cannot disable civicrm debug](https://issues.civicrm.org/jira/browse/CRM-19976)


eileenmcnaughton:
This seems consistent with other code, although it is already possible to do this using drush.
drush cvapi setting.create debugging=0

Merging as low risk & I think given the code that exists this does make it clearer

PR #460, opened 2017-07-11
Change help text

Stoob:
Change help text to accurately define the purpose of this filter, because it does not list events.  Adding another PR to list events.


Stoob:
nevermind, trying another approach

PR #461, opened 2017-07-11
add event page for PCP

Stoob:
This helps connects the PCP to the Event


Stoob:
nevermind, trying another approach

PR #462, opened 2017-07-17
CRM-20893: Port the 'civicrmtheme' module to Drupal 8

dsnopek:
* [CRM-20893: Port the 'civicrmtheme' module to Drupal 8](https://issues.civicrm.org/jira/browse/CRM-20893)

### DIFF
--- a/modules/civicrmtheme/src/Theme/CivicrmThemeNegotiator.php
+++ b/modules/civicrmtheme/src/Theme/CivicrmThemeNegotiator.php
@@ -3,6 +3,7 @@
 namespace Drupal\civicrmtheme\Theme;
 
 use Dompdf\Exception;
+use Drupal\civicrm\Civicrm;
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Session\AccountInterface;
@@ -28,14 +29,22 @@ class CivicrmThemeNegotiator implements ThemeNegotiatorInterface {
   protected $configFactory;
 
   /**
+   * The Civicrm service.
+   *
+   * @var \Drupal\civicrm\Civicrm
+   */
+  protected $civicrm;
+
+  /**
    * Constructs a CivicrmThemeNegotiator.
    *
    * @param \Drupal\Core\Session\AccountInterface $user
    *   The current user service.
    */
-  public function __construct(AccountInterface $user, ConfigFactoryInterface $config_factory) {
+  public function __construct(AccountInterface $user, ConfigFactoryInterface $config_factory, Civicrm $civicrm) {
     $this->user = $user;
     $this->configFactory = $config_factory;
+    $this->civicrm = $civicrm;
   }
 
   /**
@@ -69,7 +78,7 @@ class CivicrmThemeNegotiator implements ThemeNegotiatorInterface {
 
     // Attempt to initialize CiviCRM.
     try {
-      \Drupal::service('civicrm');
+      $this->civicrm->initialize();
     }
     catch (\Exception $e) {
       return FALSE;
@@ -85,7 +94,7 @@ class CivicrmThemeNegotiator implements ThemeNegotiatorInterface {
     $path = ltrim($route_match->getRouteObject()->getPath(), '/');
 
     // Initialize CiviCRM.
-    \Drupal::service('civicrm');
+    $this->civicrm->initialize();
 
     // Get the menu for above URL.
     $item = \CRM_Core_Menu::get($path);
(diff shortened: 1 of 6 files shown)


jackrabbithanna:
Oh wow, very cool contribution! I'll take responsibility for reviewing and testing this.

josephlacey:
I've installed and have started testing this with Drupal 8.2.x and CiviCRM 4.7.x. Everything is going swimmingly so far.  I'll report if I run into any issues.

dsnopek:
@josephlacey Thanks!

I've also had more opportunity to test this since creating the PR, and it's been working great. I've now tested it with both admin and public pages and the correct theme is used

seamuslee001:
@totten @jackrabbithanna this looks good to me i would be +1 on merging

seamuslee001:
Jenkins test this please

dsnopek:
Any updates on getting this merged? We've been using it for quite a while on our dev sites, and even teased out a bug 2 weeks (see last commit). Just looking to use a less patched module :-)

jackrabbithanna:
I'll test this this weekend, and I think that it will be good to merge after that

dsnopek:
Awesome, thanks!

modules/civicrmtheme/src/Theme/CivicrmThemeNegotiator.php
```diff
@@ -0,0 +1,102 @@
...
+    if (count($parts) > 1 && $parts[1] == 'upgrade') {
+      return FALSE;
+    }
+
+    $config = $this->configFactory->get('civicrmtheme.settings');
+    $admin_theme = $config->get('admin_theme');
+    $public_theme = $config->get('public_theme');
+
+    // @todo Should we inject the 'civicrm' service instead?
+    if ((!$admin_theme && !$public_theme) || !\Drupal::service('civicrm')) {
```
jackrabbithanna: Here, should we be using the new initialize() method of the civicrm service instead, after that other PR gets merged?

jackrabbithanna: Or maybe not...you comment there though, "Should we inject the 'civicrm' service instead....Isn't dependency injection the standard Drupal 8 practice? I'm not criticizing, I'm asking he question....

Can you help me understand what that line is doing? line 66

jackrabbithanna: Basically the question is, for me, what do you need to know with that line?  If it can't load the service, then return false, meaning civicrm couldn't be initialized? Because the constructor would return the object....so if you get no object then return false?

jackrabbithanna (commented):
Maybe if the require statement you add is included in the civicrm service initialize() method, everything will be there to avoid that unknown function error, try it out...

UPDATE:
IGNORE THIS MESSAGE POSTED ON WRONG PR CONVERSATION!!!!!

I do still want answer to the questions about the code though...

jackrabbithanna:
I have tested this module, and it works. 

I tried setting the civicrm admin to various themes, I tried setting public civicrm page to various themes...

I also made sure that when setting to "Default Theme", that it also works...

This will be good to merge, would still appreciate my one code question being answered so I understand, but other than that, this has been reviewed and tested, and its ready to be merged...

modules/civicrmtheme/src/Theme/CivicrmThemeNegotiator.php
dsnopek: Yeah, you're right, that bit is not totally useful. I ported this module by starting with the D7 code and changing it until it was a working D8 module, and so, I was converting the `civicrm_initialize()` to `\Drupal::service('civicrm')` and ended up with that bit of nonsense. :-) The service cannot be falsey (without changing how the container creates it), it can only be the service object or throw an exception.

> Basically the question is, for me, what do you need to know with that line?

So, I ported this from the D7 code, but I believe that bit is about not trying to do anything if CiviCRM isn't working, just bail early.

I've made a new commit that makes a lot more sense!

We definitely shouldn't inject the 'civicrm' service until after #495 is merged, because then civicrm will be initialized on every request, not just those that starting with `/civicrm`, which could be a performance issue. But once that is merged, it would be best practice to inject it!

dsnopek:
Replied on the one question [above](https://github.com/civicrm/civicrm-drupal/pull/462#discussion_r143489035), which GitHub is now hiding because I change that bit of code

jackrabbithanna:
Ok, I like this now, this has been reviewed and tested. @totten or @eileenmcnaughton , any additional thoughts, would you like to merge or should I do the honors?

dsnopek:
I've updated this now that #495 is merged!

jackrabbithanna:
Ok great, @colemanw  I'm pinging you here too..

Not sure who to ping, what's the standard practice you guys want?

David and I will be making and reviewing PRs, Coleman are you handling D8 merges?

jackrabbithanna:
David's last commit is good and now the Civicrm service is injected properly, thanks for adding that. This is good to go

colemanw:
@jackrabbithanna would you like to be a maintainer for this repo? If so I can give you merge rights.

PR #464, opened 2017-07-21
CRM-20937 - update to most recent templates used in core.

jmcclelland:
* [CRM-20937: civicrm engage breaks birth date field](https://issues.civicrm.org/jira/browse/CRM-20937)


eileenmcnaughton:
I'm happy to merge this based on the fact Jamie is pretty much the product owner for this code & ptp use it the most. Agree that templates need to be kept up to date (would be even better if we could avoid tpl overrides)

eileenmcnaughton:
(NB this is also functionality that is not broadly used outside of PTP I believe & which, had it been written today, would be in a separate extension)

PR #465, opened 2017-07-21
CRM-17652: Symfony conflict between Civi and Drupal 8

dsnopek:
* [CRM-17652: Symfony conflict between Civi and Drupal 8](https://issues.civicrm.org/jira/browse/CRM-17652)

### DIFF
--- a/civicrm.info.yml
+++ b/civicrm.info.yml
@@ -4,5 +4,3 @@ description: 'Constituent Relationship Management system. Allows sites to manage
 package: CiviCRM
 version: 8.x-4.7
 core: 8.x
-dependencies:
-  -  libraries:libraries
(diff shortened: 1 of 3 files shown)


dsnopek:
Now that https://github.com/civicrm/civicrm-core/pull/10694 is merged, this is the next step to a patchless install on Drupal 8!

jackrabbithanna:
Don't mean to be a stickler but can you fix the spacing before your new if() clause to meet standard Drupal coding standards. 2 less I think :) Big thanks!

jackrabbithanna:
Oh and just remove the lines you commented out please.

dsnopek:
@jackrabbithanna Ah, yes, sorry! I cleaned that bit up. Thinking about it a little more, we should probably update the docs, but I'm not super sure what to put in there since the process is still in flux, and I don't know if it makes sense for the in code docs to be constantly changing too? What do you think?

PR #466, opened 2017-07-27
CRM-20974: Remove bogus API call and just require class.api.php

dsnopek:
* [CRM-20974: Unable to install civicrm_views](https://issues.civicrm.org/jira/browse/CRM-20974)

### DIFF
--- a/modules/civicrm_views/civicrm_views.views.inc
+++ b/modules/civicrm_views/civicrm_views.views.inc
@@ -663,9 +663,7 @@ function _civicrm_views_get_custom_fields() {
             LEFT JOIN civicrm_custom_group g ON g.id = f.custom_group_id
             where g.is_active = 1";
 
-  // We make a bogus call to civicrm_api() so that it will include its requisite files
-  // for the two function calls below. Yes, this is gross — but that's what you get it if you break autoloading conventions.
-  civicrm_api(NULL, NULL, array());
+  require_once 'api/class.api.php';
 
   $custom_fields = array();
   $dao = \CRM_Core_DAO::executeQuery($query);
--- a/modules/civicrm_views/civicrm_views.views.inc
+++ b/modules/civicrm_views/civicrm_views.views.inc
@@ -663,7 +663,7 @@ function _civicrm_views_get_custom_fields() {
             LEFT JOIN civicrm_custom_group g ON g.id = f.custom_group_id
             where g.is_active = 1";
 
-  require_once 'api/class.api.php';
+  require_once 'api/v3/utils.php';
 
   $custom_fields = array();
   $dao = \CRM_Core_DAO::executeQuery($query);
--- a/modules/civicrm_views/civicrm_views.views.inc
+++ b/modules/civicrm_views/civicrm_views.views.inc
@@ -61,9 +61,9 @@ function civicrm_views_views_data() {
         'field' => 'id',
         'base' => $field['table'],
         'base field' => 'entity_id',
-        'title' => t('CiviCRM Custom Field: !title', array('!title' => $field['group'])),
-        'label' => t('CiviCRM Custom Field: !title', array('!title' => $field['group'])),
-        'help' => t('CiviCRM Custom Field: !title', array('!title' => $field['group'])),
+        'title' => t('CiviCRM Custom Field: @title', array('@title' => $field['group'])),
+        'label' => t('CiviCRM Custom Field: @title', array('@title' => $field['group'])),
+        'help' => t('CiviCRM Custom Field: @title', array('@title' => $field['group'])),
       ),
     );
 


jackrabbithanna:
@totten or @seamuslee001 , is that the right file to include?
Or should it be api/api.php ? Or what's a better way to bootstrap civi here?

Perhaps when the service is loaded that should include the necessary files:
\Drupal::service('civicrm');

Fine to fix so we can have something for the current Core Views module...But CiviCRM Entity may go into Core and deprecate Core Views....either way, doesn't work to have this working.  We may borrow some code from core Views anyway :)

jackrabbithanna:
I almost want to make that a separate issue...
Lets develop the ONE way to include necessary files, like we did for D7 with civicrm_initialize()

I think we're really close when the CiviCRM service constructor method is called:
https://github.com/civicrm/civicrm-drupal/blob/8.x-master/src/Civicrm.php

jackrabbithanna:
See https://github.com/civicrm/civicrm-drupal/blob/8.x-master/civicrm.module#L96
Notice that after the CiviCRM service is instantiated, then you can use CRM_Core_BAO_UFMatch::getContactId($account->id());

seamuslee001:
Big +1 to your comment @jackrabbithanna re one way to bootstrap. The one thing i notice that isn't in the constructor is a call to CRM_Core_ClassLoader now i don't know if that is "taken care of" through the composer process or not. Jack you would know more about class loading in D8 than I do so happy to take guidance but agree that api/api.php feels wrong

jackrabbithanna:
Maybe just try adding this:
\Drupal::service('civicrm');

instead of including the class.api.php works....

If so, maybe what we have is good already there in the CiviCRM service constructor...If that doesn't work, we'll look to update that service constructor

dsnopek:
I haven't tried just doing `\Drupal::service('civicrm')` - I'll give that a try next time I'm ready to test this!

I know this really isn't the place to discuss this (I'll make a JIRA ticket at some point) but it bothers me that that service constructor has so many side-effects. Generally, constructing a services should do nothing but initialize the object. Since a service might need to be constructed just to resolve a dependency, it doesn't necessarily mean any work is actually about to be done with the service.

It would feel better to me if this magic incantation was actually `\Drupal::service('civicrm')->initialize()` because then you're actually calling a method to cause all those side-effects.

jackrabbithanna:
@dsnopek , I think you are right about the initialize() function. We also may want to add different methods to the service later too for specific purposes, possibly without the general overhead in the constructor... Probably best to make this a separate issue and work on it and discuss it there....

dsnopek:
@jackrabbithanna Issue created! https://issues.civicrm.org/jira/browse/CRM-21093

dsnopek:
Just a note that with a custom field, my changes here crash with a fatal error:

```
Error: Call to undefined function _getStandardTypeFromCustomDataType() in _civicrm_views_get_custom_fields()
```

I tried switching to `\Drupal::service('civicrm')` but that doesn't work either. There must be some other magic incantation to load the necessary code...

jackrabbithanna:
Should we add your require statement to the initialize() method of the CiviCRM Service...

I'll start looking around for what needs to be included to get the undefined error fixed

jackrabbithanna:
Maybe if the require statement you add is included in the civicrm service initialize() method, everything will be there to avoid that unknown function error, try it out...

Merging this PR will likely depend on merging #495

dsnopek:
Per my [last comment](https://github.com/civicrm/civicrm-drupal/pull/466#issuecomment-331611920), neither my original fix nor initializing the CiviCRM service actually allow the Views code for custom fields to work. My original code will stop the site from crashing just by enabling civicrm_views, but as soon as you add a custom field, it'll crash again. So, this still needs some work!

jackrabbithanna:
I understand I'm thinking we find the right thing to include, and put it in the initialize() method of the Civicrm service, or another method in that service....

jackrabbithanna:
Here's what the API does to initialize itself:
https://github.com/civicrm/civicrm-core/blob/master/api/v3/utils.php#L37

Maybe that's the stuff that made the "bogus api call" work for this before....

David, how about adding to the initialize() method of the Civicrm service 

require_once 'CRM/Core/ClassLoader.php';

and testing

If that doesn't work maybe we add 
CRM_Core_ClassLoader::singleton()->register();

and testing...

@colemanw do you have thoughts on this?

dsnopek:
I finally got to the bottom of this! The file that needs to be required to make this code actually work with a custom field is:

```
require_once 'api/v3/utils.php';
```

Also, while testing it, I also added a commit to make it correctly show the names of custom fields in the Views UI. The problem is that this code was using `t()` with arguments starting with an exclamation point, which was removed at some point in Drupal 8's development cycle before 8.0.0 came out.

Anyway, I've actually tested this code to work!

jackrabbithanna:
Ok this is good news, I only wonder if we should put these require statements in the CiviCRM service...Just pondering that bit...basically should we start a standard where the CiviCRM service is used to support any necessary require statements in Core or in other submodules, and then use a method on the service to do the require.  Thoughts?

seamuslee001:
I like the way your thinking Mark, so the thing is i believe is that `_getStandardTypeFromCustomDataType` is located in the api_v3_Utils.php what supprises me a bit but maybe understandable is that if CRM_Core_DAO class is loaded why doesn't api/v3/Utils.php get loaded as well? what does that mean for the other API classes? i think the more it happens in the CiviCRM service the simpler it would be for custom modules etc to implement CiviCRM integration

jackrabbithanna:
Ok so the question becomes, should we put this into the initialize() method...or into a new method, just for this require statement...

Hopefully as the Core submodules get ported, we'll find a pattern, and can group them appropriately, as necesssary, with the intention of freezing the "api" of this service before the Drupal 8 version is officially released?

jackrabbithanna:
I've tested the code in the PR and can verify it works.  After we sort out where the require statements will end up I think this will be good

jackrabbithanna:
@dsnopek I think for now, let's put the require statement you have here, in the initialize() method of the CiviCRM service...
and then the invoking of that method in civicrm_views_views_data() on line 7 should give you what you need for the call to _civicrm_views_get_custom_fields()

If there becomes others that need to be added as development continues and we find patterns, we'll think about additional methods in the Civicrm service. 

How does that sound?

dsnopek:
I dunno... If the Drupal 7 version of the CiviCRM module doesn't ensure that file is loaded after calling `civicrm_initialize()`, maybe it isn't necessary for Drupal 8 either? Since it's only that very specific Views hook that needs that file, and that hook is only invoked on cache clear (ie. rarely), it seems like overkill to _always_ require that file since most of the time we don't need it. Keeping unnecessary code out of memory and the OPCache seems like a good thing?

jackrabbithanna:
A separate method in the service would suffice too

dsnopek:
I'm not really sure about that either. Do we really know how often people need to require this file? Is it a frequent enough need that we should create a new API on the service for it? So far, there's just this Views code that needs it. Since Drupal 7 didn't require this file in it's `civicrm_initialize()` function and somehow got along fine, I suspect that not much else will need this... My inclination is just to require the file in this one spot, and wait see if some other code needs to require it too, and then consider making an API for it.

jackrabbithanna:
Well the thing is we probably don't know how often people need to require this file.  I'm not too keen to argue about it here on this issue, been I've been in this ecosystem long enough to be expect to be surprised occasionally. We have yet to see many other modules developed for Civi + D8, and we'll see what comes, especially out of Webform CiviCRM....(for CiviCRM Entity I try never to do anything except by CiviCRM API, not using any DAO/BAO etc.. functions)

I'd really like to see the require in the service, and instead of debating it on an issue to get CiviCRM Views going, I will likely create another PR for it, and open up the discussion there. 

Merging #466 to get the ball down the field

dsnopek:
Ok, cool, thanks! Sorry for being difficult. My tendancy is to try to avoid prematurely creating APIs (like public methods on a service), because then you have to maintain them. :-)

Maybe the lookup that civicrm_views is doing could be done via `civicrm_api3()` rather than calling that utility function directly? I don't totally understand what it's doing, I was just trying to fix the fatal.

dsnopek:
And thanks for merging! :-)

jackrabbithanna:
Yeah I think the service having an API, and what patterns that API needs to cover is just a topic we could debate for a while, and I don't want to stop progress on getting "something that works" needlessly

PR #468, opened 2017-08-07
CRM-21016: Allow Drupal views to be used for Notes related to contributions

monishdeb:
Not only contributions but now note fields will be available for Participant and Relationships.

---

 * [CRM-21016: Allow Drupal views to be used for Notes related to contributions](https://issues.civicrm.org/jira/browse/CRM-21016)


monishdeb:
@colemanw can you please test this PR

colemanw:
Code looks good. Tested in views and works great!

PR #469, opened 2017-08-10
Add Contributions relevant to the Participant record.

JohnFF:
Add Contributions relevant to the Participant record so that they can be used in Drupal Views related to Participants.


eileenmcnaughton:
this seems fine & as it is just adding metadata it feels very safe

PR #470, opened 2017-08-10
Fix issue CRM-20292 - clear cache breaks Views of custom fields.

JohnFF:
Added views_invalidate_cache all to drush cc civicrm. This prevents an issue where clearing cache would prevent civicrm custom fields from appearing in views after drush cc all called.

---

 * [CRM-20292: Drush cc all clears custom fields from Drupal Views](https://issues.civicrm.org/jira/browse/CRM-20292)

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -829,6 +829,12 @@ function drush_civicrm_cacheclear() {
     return;
   }
 
+  // Without this, custom fields disappear from views that use them.
+  // Check that the views module is enabled by testing the presence of a function.
+  if (function_exists('views_invalidate_cache')) {
+    views_invalidate_cache();
+  }
+
   drush_log(dt('The CiviCRM cache has been cleared.'), 'ok');
 }
 


jackrabbithanna:
This looks good, I recommend merging. @eileenmcnaughton are we using the 4.6 label for PRs for the Drupal module 4.7 vs. 4.6? I think this would be good for both versions.   @JohnFF Let me get an answer from Eileen or somebody on my question, you may need to make a PR against the 4.6 branch of this repo to get it into the next version of 4.6....

JohnFF:
Sadly, I am pulling this back, as it doesn't seem to fully address the issue as I'd hoped.

eileenmcnaughton:
@jackrabbithanna I think github has a bot that adds the labels - we don't do that manually. Note that if the submitter doesn't have an interest in putting to 4.6 you can cherry-pick yourself & create the PR. I don't know who is supposed to merge it in that case since I guess you wouldn't be able to then

JohnFF:
@eileenmcnaughton @jackrabbithanna I am withdrawing this PR as it has a configuration dependency that the ideal solution would not need.

PR #471, opened 2017-08-25
Apply Environment Indicator colours to CiviCRM menu also.

xurizaemon:
[CRM-21103: Make CiviCRM menu respond to Drupal Environment Indicator module if present](https://issues.civicrm.org/jira/browse/CRM-21103)

image

### DIFF
--- a/civicrm.info
+++ b/civicrm.info
@@ -6,6 +6,8 @@ core = 7.x
 project = civicrm
 php = 5.3
 
+scripts[] = js/civicrm-drupal.js
+
 files[] = civicrm.module
 files[] = civicrm.install
 files[] = civicrm_user.inc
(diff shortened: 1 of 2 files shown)


colemanw:
IMHO this ought to be contrib not core.

colemanw:
Also note that there is already a civicrm-drupal.js file in the civicrm repo, so we'd want to be more consistent about that.

xurizaemon:
Fair enough, will release as an extension. Thanks for review!

xurizaemon:
I did see crm.drupal.js (I think) but not civicrm-drupal.js, and I *thought* about adding it there. However, I wasn't clear on why a .drupal.js was in core in the first place ... so didn't want to add to that! It felt a better fit in civicrm-drupal repo. I agree it makes much more sense in an extension.

xurizaemon:
Should be available from CiviCRM.org once it's reviewed - https://civicrm.org/extensions/environment-indicator-integration

PR #472, opened 2017-08-27
Fixed CRM-20292: removed permission checks for custom data caching in views

JohnFF:
Having the permission check breaks views with custom field, (they are broken even in the admin view), unless there is an ACL in place if Drush CC is used on the command line.

Saving any other view as an administrator fixed this.

---

 * [CRM-20292: Drush cc all clears custom fields from Drupal Views](https://issues.civicrm.org/jira/browse/CRM-20292)

### DIFF
--- a/modules/views/civicrm.views.inc
+++ b/modules/views/civicrm.views.inc
@@ -327,7 +327,8 @@ function civicrm_views_custom_data_cache(&$data, $entity_type, $group_id, $sub_t
     return;
   }
   // From http://forum.civicrm.org/index.php/topic,17658.msg73901.html#msg73901, CRM-7860.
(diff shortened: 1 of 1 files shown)


JohnFF:
https://issues.civicrm.org/jira/browse/CRM-20292

seamuslee001:
Jenkins test this please

seamuslee001:
Jenkins re test this please

seamuslee001:
I had a chat with John FF about this and I agree with the change, It makes sense that this is the correct change and that it stabilizes behavior. @totten @eileenmcnaughton i believe this should get merged

JohnFF:
@totten @eileenmcnaughton any news? This has been ready for merge for around 2 months, is a 1 parameter change to ensure consistent behaviour.

eileenmcnaughton:
Ok - I'll merge based on @seamuslee001 review

PR #473, opened 2017-08-27
(NFC) Lint civicrm.drush.inc

seamuslee001:


totten:
Nice. To my reading, it looks safe (only see whitespace and docblock changes).

PR #474, opened 2017-08-28
(NFC) Lint civicrm.config.php.drupal

seamuslee001:


seamuslee001:
Jenkins test this please

totten:
Yes! Fixing the syntax of this file feels so right, but nothing can be that right, so it must be wrong. But fie on that! Let's fix it anyway!

PR #477, opened 2017-08-28
Lint CiviCRM Rules module

seamuslee001:


seamuslee001:
Jenkins test this please

seamuslee001:
Jenkins test this please

seamuslee001:
Jenkins test this please

modules/civicrm_rules/civicrm_rules.module
```diff
@@ -33,13 +33,6 @@
  *
  */
 
-/*
```
seamuslee001, line 36: Movement of this is as per "The hook is called every time the module is enabled. It should be implemented in the module's .install file. The implementation can rely on the .module file being loaded."

From http://www.drupalcontrib.org/api/drupal/drupal%21modules%21system%21system.api.php/function/hook_enable/7

eileenmcnaughton:
I eyeballed these & the only functional change is moving the enable function to the install file, which makes sense

PR #478, opened 2017-09-06
CiviCRM / Drupal Rules: Show all groups

xurizaemon:
CRM-21156

---

 * [CRM-21156: CiviCRM Rules does not show all available Groups](https://issues.civicrm.org/jira/browse/CRM-21156)

### DIFF
--- a/modules/civicrm_rules/civicrm_rules_utils.inc
+++ b/modules/civicrm_rules/civicrm_rules_utils.inc
@@ -42,7 +42,13 @@
  */
 function _civicrm_get_groups() {
   civicrm_initialize();
-  $groupInformation = civicrm_api('group', 'get', array('version' => 3));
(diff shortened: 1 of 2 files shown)


modules/civicrm_rules/civicrm_rules_utils.inc
```diff
@@ -42,7 +42,13 @@
  */
 function _civicrm_get_groups() {
   civicrm_initialize();
-  $groupInformation = civicrm_api('group', 'get', array('version' => 3));
+  $groupInformation = civicrm_api('group', 'get', array(
+    'version' => 3,
```
jitendrapurohit, line 46: Tested this and it works fine. Maybe, its good to write `civicrm_api3()` instead of passing `version` as a param?

PR #487, opened 2017-09-16
Run Civilint against the rest of views_handler and views plugin files

seamuslee001:
ping @eileenmcnaughton @totten @monishdeb

### DIFF
--- a/modules/views/civicrm/civicrm_handler_filter_country_multi.inc
+++ b/modules/views/civicrm/civicrm_handler_filter_country_multi.inc
@@ -21,22 +21,23 @@
    | GNU Affero General Public License or the licensing of CiviCRM,     |
    | see the CiviCRM license FAQ at http://civicrm.org/licensing        |
    +--------------------------------------------------------------------+
+ */
 
-   **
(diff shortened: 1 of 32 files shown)


totten:
Regarding the change from `function foo()` => `public function foo()`: I agree this should be safe, but (a) folks have made mistakes on that calculation before and (b) AFAIK there's no test-coverage which would load these files. Did a few things to check these risks:

 * [Played around](https://gist.github.com/totten/936826b442e20d492cfc795f91369a11), I can't find a likely situation where this would cause a break.
 * I wrote small throw-away script which simply does `require_once` on each `*.inc` file (then executed with `cv scr`). There are a couple files that throw pre-existing errors (eg `calendar_plugin_*.inc`), but this PR doesn't change the list of loading/non-loadable files.

seamuslee001:
Also from here @totten http://www.php.net/manual/en/language.oop5.visibility.php  "Class methods may be defined as public, private, or protected. Methods declared without any explicit visibility keyword are defined as public." so the fact they didn't have anything before meant php was treating them as public

seamuslee001:
@totten any further feelings on this?

PR #488, opened 2017-09-16
Run civilint against civicrm_handler_filter_state_multi.inc

seamuslee001:
This one will need some testing done to it, it will need to be checked that the JS output is the same as previous the changes should be fine

### DIFF
--- a/modules/views/civicrm/civicrm_handler_filter_state_multi.inc
+++ b/modules/views/civicrm/civicrm_handler_filter_state_multi.inc
@@ -82,7 +82,7 @@ class civicrm_handler_filter_state_multi extends views_handler_filter_in_operato
         drupal_add_js(array('civicrm_views' => array('states' => $state_html)), 'setting');
         $c_object = json_encode($state_html);
       }
-    } // end foreach
+    }
 
     // country filter exists, remove options, add js
     if (!empty($country_field)) {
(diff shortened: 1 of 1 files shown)


modules/views/civicrm/civicrm_handler_filter_state_multi.inc
```diff
@@ -89,63 +88,65 @@ class civicrm_handler_filter_state_multi extends views_handler_filter_in_operato
...
-      $fid=md5($country_field);
-      $js = '(function ($) {
-               Drupal.behaviors.civicrm_views_' . $fid . ' = {
+      $fid = md5($country_field);
+      $js = "(function ($) {
+               Drupal.behaviors.civicrm_views_{$fid}= {
                  attach: function (context, settings) {
-                   $("#edit-' . $country_field . '").change( function() {
-                     var stateList = \'<option value="All">&lt;Any&gt;</option>\';
+                   $(#" . '"' . "edit-{$country_field}" . '"' . ").change( function() {
```
colemanw: This is not the same. Note the hash mark is now outside the quotes.

colemanw (changes requested):

modules/views/civicrm/civicrm_handler_filter_state_multi.inc
seamuslee001: Thanks @colemanw

seamuslee001:
@colemanw should be fixed up now

PR #491, opened 2017-09-18
Run civilint against civitest.module.sample

seamuslee001:


civitest.module.sample
```diff
@@ -165,197 +170,174 @@ function civitest_civicrm_buildAmount(
...
-    }
-
-    foreach( $contactIDs as $cid ) {
-      foreach( $tokens['views'] as $token ) {
-        $values[$cid]["views.{$token}"] = $_cache[$token];
-      }
-    }
-  }
-}
+//function civitest_civicrm_tokenValues(&$values,
```
totten, line 317: Why is this changed to being commented out?

(TBH, I don't think it's particularly problematic -- since the main idea of the file is to capture a bunch of code-snippets that can be copied/edited, and comments do that just as well as code... but it feels little arbitrary to comment out `hook_civicrm_tokenValues` with `//` but not anything else.)

civitest.module.sample
```diff
@@ -1510,47 +1509,48 @@ function civitest_civicrm_searchColumns( $objectName, &$headers,  &$values, &$se
...
-function civitest_civicrm_dupeQuery( $obj, $type, &$data ) {
+function civitest_civicrm_dupeQuery($obj, $type, &$data) {
   //don't run these during user account/contact creation
-  if( $obj->noRules || $type != 'table')
+  if ($obj->noRules || $type != 'table') {
     return;
-
-  if ( $obj->name=='My Dedupe Rule Group Name' ) {
+  }
+  if ($obj->namei == 'My Dedupe Rule Group Name') {
```
totten, line 1538: `name` vs `namei`?

civitest.module.sample
totten: OK, so there's a couple more functions where the PR comments them out entirely. Same question applies to those.

civitest.module.sample
seamuslee001: Ooops

civitest.module.sample
seamuslee001: the reason is that Civilint was detecting 2 functions with the same name and i figured each had enough "good" content to warrent both of them being there so just commented one out.

seamuslee001:
@totten @eileenmcnaughton i'm also wondering about the usefulness of this file in particular given the excellent dev-docs we now have, Should we rather than doing it like this we actually remove this file all together? noting it is suffixed with .sample

civitest.module.sample
totten: OK, that makes sense. 👍

totten:
Anecdotally, I've neither referred people to `civitest.module.sample` nor noticed others refer to it. My sense is that it was an early iteration on how to document hooks (gradually forgotten when the wiki pages became standard-practice). After searching Google and grepping `universe`, there seem to be:

 * A couple references from dev-training discussions circa 2009-2010
 * A couple references copied into `civicrm-dev-docs` (`civicrm-dev-docs/docs/hooks/hook_civicrm_dashboard.md` and `civicrm-dev-docs/docs/hooks/hook_civicrm_emailProcessorContact.md`).

+1 for removing it and for updating those two files in `civicrm-dev-docs`.

seamuslee001:
I'm going to close this in favor of https://github.com/civicrm/civicrm-drupal/pull/492 as i think that is the more correct idea here given the dev-docs should be the single point for examples of hooks

PR #493, opened 2017-09-19
Run Civilint against civicrm_engage module

seamuslee001:


seamuslee001:
ping @eileenmcnaughton @totten @colemanw i believe this is very simple

PR #494, opened 2017-09-25
Need to include fields in order by  in select distinct to work

beto-aveiga:
Hi! I'm coming from here https://issues.civicrm.org/jira/browse/CRM-21132
I don't have an account approved yet so I can't post anything there.
A select distinct with order by needs to include the fields in the order by.
Hope that helps.


seamuslee001:
Jenkins ok to test

seamuslee001:
Just noting the reporter has confirmed on the JIRA ticket this works. Unit tests have also passed. The change is just adding an extra column into the select. I believe that this should now be merged @totten @eileenmcnaughton @colemanw

PR #495, opened 2017-09-26
CRM-21093: Move CiviCRM initialization from service constructor to method.

dsnopek:
See https://issues.civicrm.org/jira/browse/CRM-21093

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -94,7 +94,7 @@ function civicrm_entity_extra_field_info() {
  * to the user display.
  */
 function civicrm_user_view(array &$build, \Drupal\user\UserInterface $account, \Drupal\Core\Entity\Display\EntityViewDisplayInterface $display, $view_mode) {
-  \Drupal::service('civicrm');
+  \Drupal::service('civicrm')->initialize();
 
   // We need the $contact_id so that we know what data to pull out of Civicrm.
   // And we need the contact_id of the current user ($current_contact_id) so
@@ -177,7 +177,7 @@ function civicrm_user_view(array &$build, \Drupal\user\UserInterface $account, \
  * based on the $contact_id's contact type.
  */
 function _civicrm_get_profiles($contact_id) {
-  \Drupal::service('civicrm');
+  \Drupal::service('civicrm')->initialize();
   $profiles = array();
   $ctype = \CRM_Core_DAO::getFieldValue('CRM_Contact_DAO_Contact', $contact_id, 'contact_type');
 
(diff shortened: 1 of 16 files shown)


jackrabbithanna:
The PR looks good to me, but it failed the tests....Although when I look into why it failed, the console output was "Build step 'Publish xUnit test result report' changed build result to FAILURE"

@bgm or @totten would you'll have an idea what that means?

jackrabbithanna:
Other than that I think that this should be merged

seamuslee001:
I believe this looks ok to me but haven't tested @jackrabbithanna re tests, Tests are currently not enabled for Drupal 8 because the PR unit test box is php5.3 not php5.4 (as a min).  In the console output it has "PR test not allowed for 8.x-master"

jackrabbithanna:
Ok it looks good so far, but there is one more place you need to fix.

I'm gettnig PHP Fatal error:  Class 'CRM_Core_BAO_UFGroup' not found in /var/www/html/drupal83/modules/civicrm-drupal/src/Form/UserProfile.php on line 91

Going to the user view page, and a error on the user edit page...

need to add 
\Drupal::service('civicrm')->initialize();

To the UserProfile class, in src/Form/UserProfile.php

jackrabbithanna:
I also think you will need to add it to the __construct() of the CivicrmContoller Class in src/Controller/CiviCRMController.php

jackrabbithanna:
I'm getting this error, when clearing Drupal caches:
Call to undefined function civicrm_api3_create_error() in /var/www/html/drupal83/vendor/civicrm/civicrm-core/Civi/API/Kernel.php on line 413

Was this fixed in another issue, and my test instance is behind?

Every Drupal page will WSOD after that, until I go to /civicrm

jackrabbithanna:
I remember now, the cache clear WSOD is caused by civicrm_views, which you were fixing here:

https://github.com/civicrm/civicrm-drupal/pull/466/files

Perhaps we should add the require statement below:
require_once 'api/class.api.php';

To the initialize() function of the service, and then instantiate the service and run the initialize() method on it in PR #466

dsnopek:
> I'm gettnig PHP Fatal error: Class 'CRM_Core_BAO_UFGroup' not found in /var/www/html/drupal83/modules/civicrm-drupal/src/Form/UserProfile.php on line 91

Oops! Thanks, I missed one. :-) I searched for other instances for the comment about not doing anything with the Civicrm service and found one place I forgot to clean it up, but not more needed functional changes.

I've pushed a couple new commits for those things!

> I also think you will need to add it to the __construct() of the CivicrmContoller Class in src/Controller/CiviCRMController.php

I don't think so, actually, because `$civicrm->invoke()` will automatically initialize the service if it already hasn't been.

> I remember now, the cache clear WSOD is caused by civicrm_views, which you were fixing here #466 

The fix in #466 doesn't actually work yet :-( Let's get this one merged (once it all looks good, of course!) and figure that one out on that PR

jackrabbithanna:
Ok, @totten, @eileenmcnaughton @colemanw I've tested this PR and reviewed the code. It is all good from my perspective, and other PRs are depending on this one to move forward.  I'm requesting this get merged.

dsnopek:
Woohoo, thanks! :-)

PR #496, opened 2017-10-11
CRM-21292: Convert all SafeMarkup call to non-deprecated substitutes

dsnopek:
See https://issues.civicrm.org/jira/browse/CRM-21292


dsnopek:
Woohoo! Thanks for the quick review and merge :-)

PR #500, opened 2017-10-14
CRM-19995 - Fix missing 's' in array name

olivierh65:
Without the 's', the variable is not an array, and give this error : 
Notice : Undefined offset: 5 dans _civicrm_member_roles_sync() (ligne 587 dans /xxxxxxx/sites/all/modules/civicrm/drupal/modules/civicrm_member_roles/civicrm_member_roles.module).

---

 * [CRM-19995: Notice : Undefined offset: 5 dans _civicrm_member_roles_sync\(\)](https://issues.civicrm.org/jira/browse/CRM-19995)


jackrabbithanna:
Yeah olivierh65 is right, unsetting $membership does nothing that is persistant beyond the foreach() scope

This looks good.

jackrabbithanna:
Jenkins test this please

PR #501, opened 2017-10-23
CRM-21064: Add composer.json

dsnopek:
This adds a composer.json to the Drupal module so it's installable via composer

See https://issues.civicrm.org/jira/browse/CRM-21064

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,35 @@
+{
+  "name": "civicrm/civicrm-drupal",
+  "description": "Open source constituent relationship management for non-profits, NGOs and advocacy organizations.",
+  "type": "drupal-module",
+  "license": "AGPL-3.0",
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
@colemanw  or @totten , this should be good to merge, but make sure you have the license you want and the author information is what you desire

seamuslee001:
@dsnopek @jackrabbithanna do we need to specify a requirement of drupal/drupal or is that just implicit? 

Agree this is good

dsnopek:
@seamuslee001 Only if we want to give a version constraint (like, saying it'll only work with Drupal 8.4.0 and greater, or something) which isn't necessary here. Otherwise, it's implied!

dsnopek:
Woohoo! Thanks :-)

PR #502, opened 2017-10-25
Allow Drupal 8 installations with the vendor directory outside the web root

dsnopek:
It's considered best practice with composer to keep the vendor directory outside of the web root, however, this isn't how Drupal 8 is setup by default (in consideration of less technical users and common shared hosting platforms where this isn't possible).

A number of people following the composer instructions I posted on my blog tried installing civicrm with the vendor directory outside the web root and had problems!

This PR will allow the 'civicrm' module to be installed when in this arrangement - however, it doesn't fix all the problems, but the remaining problems are outside of the Drupal module and will need to be addressed in my instructions.

I also took this opportunity to change some of the comments and messages in the civicrm.install file that reference installing CiviCRM in the libraries directory, because that's not how this works any more!

### DIFF
--- a/civicrm.install
+++ b/civicrm.install
@@ -78,7 +78,7 @@ function civicrm_requirements($phase) {
       'title' => 'CiviCRM location',
       'value' => NULL,
       'severity' => REQUIREMENT_ERROR,
-      'description' => 'CiviCRM must be downloaded into the libraries folder.',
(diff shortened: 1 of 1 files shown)


dsnopek:
@colemanw Can you check this one out too? It's a simple change, and is one step closer to using composer according to "best practices". Thanks!

dsnopek:
Thanks!

PR #503, opened 2017-10-31
CRM-21371 - An api error when calling drush cvapi should be treated a…

eileenmcnaughton:
…s a drush error

Also removes an unused var & defines a used var

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -29,12 +29,7 @@
  *
  * @package CRM
  * @copyright CiviCRM LLC (c) 2004-2017
- * $Id$
- *
  */
-
-// $Id: example.drush.inc,v 1.5 2009/09/30 13:15:54 weitzman Exp $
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
My only concern here is that failing api calls will become 'chattier' and people may have been making use of their existing tendency to fail quietly

eileenmcnaughton:
@mlutfy are you up to merge this - @jmcclelland tested it & commented on JIRA so it has already had some review (I would also contend it's low risk - although I guess there is the question of whether new 'chattiness' would upset some people?)

mlutfy:
I did a bit of testing and seems good to me. I agree with Jamie, the exit code is probably one of the best parts of this.

eileenmcnaughton:
Right - it was the lack of exit code that caused me to dig into it

eileenmcnaughton:
(we weren't noticing fails)

PR #504, opened 2017-10-31
CRM-21376: Make profiles on user page in Drupal 8 closer to how they looked in Drupal 7

dsnopek:
See https://issues.civicrm.org/jira/browse/CRM-21376

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -20,17 +20,28 @@ function civicrm_page_attachments(array &$page) {
   CRM_Core_Resources::singleton()->addCoreStyles();
 
   // Merge CiviCRM stuff into HTML header.
-  if ($region = CRM_Core_Region::instance('html-header', FALSE)) {
-    CRM_Utils_System::addHTMLHead($region->render('', FALSE));
+  if ($region = \CRM_Core_Region::instance('html-header', FALSE)) {
+    \CRM_Utils_System::addHTMLHead($region->render(''));
   }
 
-  $headers = \Drupal::service('civicrm.page_state')->getHtmlHeaders();
+  /** @var \Drupal\civicrm\CivicrmPageState $page_state */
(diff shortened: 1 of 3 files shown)


dsnopek:
Found some problems with this in further testing! I'll try to push a commit to fix today or tomorrow. Sorry!

dsnopek:
Ok, I just pushed a new commit that fixes the bug I found yesterday!

FYI, this PR changes how assets are added to the page: the problem was that CiviCRM assets were only being added on pages under /civicrm, but not other pages (like the user view page where the CiviCRM profiles are shown). This PR makes it so that assets are always added if CiviCRM has been initialized and any were added to `\CRM_Core_Region::instance('html-header')` _and_ it makes sure that the core styles are added on all pages, including those not under /civicrm (while full core resources are added on /civicrm pages).

This matches the behavior of the Drupal 7 module (although, the specifics of the implementation are quite different because Drupal 8 is quite different) where I took a ton of inspiration. So, while the changes may look big relative to what was in the Drupal 8 module previously, they actually move things closer to the way the Drupal 7 module worked.

jackrabbithanna:
I've tested this code, and profiles show on the user view page, and very similar to as on Drupal 7...

I'm testing now with Drupal 8.4.2 and CiviCRM 4.7.27

As far as code goes, the way the CSS and JS resourses are added in hook_page_requirements should be an improvement in general to the way those resources are loaded, and viewing the profile via a twig template is great, and allows for standard Drupal theming of the profile display

I think this is good....I'll give a day or so to see if anyone else wants to review and/or test this code, but should be good to merge

dsnopek:
@jackrabbithanna Thanks for testing! :-)

jackrabbithanna:
I can't think of any reason not to merge this PR...so its going in the can
merging #504

dsnopek:
Woohoo, thanks!

PR #505, opened 2017-11-14
Allow adding user entity fields to views, not just columns on the 'users' table

dsnopek:
Currently, if you make a relationship in Views back to 'Drupal Users' it will only let you add fields for columns on the 'users' table that have Views integration. However, adding this key will tell Views that this is a relationship to the 'user' entity type, which will allow you to add any fields that are on the 'user' entity.

### DIFF
--- a/modules/civicrm_views/civicrm_views.views.inc
+++ b/modules/civicrm_views/civicrm_views.views.inc
@@ -471,6 +471,7 @@ function _civicrm_views_add_implicit_joins(&$data) {
       'users' => array(
         'left_field' => 'uid',
         'field' => 'uf_id',
+        'entity_type' => 'user',
         'extra' => array(
           array(
             'field' => 'domain_id',


dsnopek:
Ah, crap, nevermind - this didn't actually work :-/ I was testing in the wrong tab... Still trying to figure this out!

dsnopek:
I spent several hours on this yesterday, and got no where! Joining against 'users_field_data' rather than 'users' got all the fields, but still errored out on rendering because Views couldn't determine the entity type. I think the root problem is that those implicit joins won't work, and the relationship handler needs to be improved to join all the way to the Drupal 'users' and 'users_field_data' tables, because it seems that Views will defer to the relationship to determine the entity type? In any case, this started turning into some really big changes to how the Views integration is done which were super tricky and it started to seem like it wasn't worth it, because eventually we want Views integration via civicrm_entity anyway.

I'm going to close this PR for now...

jackrabbithanna:
Yeah @dsnopek , I really loathe those implicit joins....maybe we can work to not have that, and make people add normal relationships like the rest of the Drupal world.

PR #509, opened 2018-01-05
CRM-21628 Allows selection of State given a Country in the Profile creation/edit form

yaazkal:
This fixes CRM-21628 https://issues.civicrm.org/jira/browse/CRM-21628

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -16,8 +16,8 @@ function civicrm_page_attachments(array &$page) {
     return;
   }
 
-  // Always add core styles (even if not on a /civicrm page).
-  CRM_Core_Resources::singleton()->addCoreStyles();
+  // Always add core resources (even if not on a /civicrm page).
+  CRM_Core_Resources::singleton()->addCoreResources();
 
   // Merge CiviCRM stuff into HTML header.
   if ($region = \CRM_Core_Region::instance('html-header', FALSE)) {


civicrm.module
```diff
@@ -16,8 +16,8 @@ function civicrm_page_attachments(array &$page) {
     return;
   }
 
-  // Always add core styles (even if not on a /civicrm page).
```
seamuslee001, line 19: @jackrabbithanna @dsnopek I feel like these lines should stay in but the new L19 and L20 should be added thoughts?

dsnopek:
I'll need to dig into this deeper, but my gut reaction is that this change seems not quite right. This would mean that we're adding the CiviCRM core Javascript on every page where CiviCRM has been initialized. This could slow down page loads unnecessarily (what if the Javascript isn't even being used?), but also, if I recall correctly, the Drupal 7 civicrm module has this same behavior: it only adds the CSS always, and it's up to the specific CiviCRM code generating the page to add the Javascript resources if they're needed.

GIven that the support for adding resources to the page in Drupal 8 is only half implemented at this point (ex. we can't add JS to the footer, etc) and a little buggy, I'd suspect that the culprit might actually be that code somewhere.

I'll try to find some time to look at this deeper over the weekend or early next week!

yaazkal:
@dsnopek thanks !
What I see is that the resources are actually only loaded on the user registration form or in the edit profile form, so I was really little conserned about performance.

I can see that the `addCoreResources()` method has a comment telling that the there should be more functions but still are not existent:

```
TODO: Separate the functional code (like addStyle/addScript) from the policy code
(like addCoreResources/addCoreStyles)
```
I do agree there should be a better method, even try to implement something like libraries in the module to attach only the resrouces needed, but at this moment and taking in mind this module is still under active development is an easy way to allow the functionality to work, anyway is gonna need a refactoring.

dsnopek:
> What I see is that the resources are actually only loaded on the user registration form or in the edit profile form, so I was really little conserned about performance.

The change in this PR will cause the CiviCRM core Javascript to be added on any page where CiviCRM is initialized. Now, on your site, that might only be the registration form and edit profile form, and maybe the public pages for events or contributions or something.

However, we just launched a client site with a block that displays some information about the current user's membership, which is shown on every page. That block initializes CiviCRM, and so this would mean that many pages would go from having no Javascript at all, to having the CiviCRM core Javascript added, even though they don't actually need this for anything.

Looking at the Drupal 7 module, it's only adding the CiviCRM core styles on every page, and adding both styles and javascript only on `/civicrm` pages:

```
  if (arg(0) == 'civicrm') {
    // Add our standard css & js
    CRM_Core_Resources::singleton()->addCoreResources();
  }
  else {
    CRM_Core_Resources::singleton()->addCoreStyles();
  }
```

https://github.com/civicrm/civicrm-drupal/blob/7.x-master/civicrm.module#L38

And the way that it gets the Javascript added on the profile pages, is by calling `CRM_Core_Resources::singleton()->addCoreResources()` just when rendering the profile pages:

1. `civicrm_register_data()` and `civicrm_form_data()` both call `civicrm_add_jquery()`, see https://github.com/civicrm/civicrm-drupal/blob/7.x-master/civicrm.module#L546 and https://github.com/civicrm/civicrm-drupal/blob/7.x-master/civicrm.module#L605 respectively
2. `civicrm_add_jquery()` calls `CRM_Core_Resources::singleton()->addCoreResources();`, see https://github.com/civicrm/civicrm-drupal/blob/7.x-master/civicrm.module#L812

So, I think we should do something similar in the Drupal 8 version of this module. I think that means calling that function in `civicrm_user_view()` and `civicrm_form_user_register_form_alter()` but I haven't tested it yet. The result would be only adding the Javascript specifically on user registration and edit profile -- not every page that CiviCRM is initialized on.

dsnopek:
I just created an alternate PR based on my comment above! It works in my testing. See https://github.com/civicrm/civicrm-drupal/pull/512

yaazkal:
oh @dsnopek so what you said makes more sense.  I vote that this patch should be descarted then. Do you want me to close this PR?

dsnopek:
Sure, if you agree and if my PR works in your testing, then let's close this one!

jackrabbithanna:
Looks like @dsnopek put in some good analysis of this, and the original contributor agreed this PR should be closed in favor of his...closing

PR #510, opened 2018-01-06
CRM-21381 Prevents contact duplication

yaazkal:
See https://issues.civicrm.org/jira/browse/CRM-21381

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -235,7 +235,7 @@ function _civicrm_get_profiles($contact_id) {
 function civicrm_form_user_register_form_alter(&$form, &$form_state, $form_id) {
   $civicrm = \Drupal::service('civicrm');
   $civicrm->initialize();
-  $html = \CRM_Core_BAO_UFGroup::getEditHTML(NULL, '', 1, TRUE, FALSE, NULL, FALSE, $civicrm->getCtype());
+  $html = \CRM_Core_BAO_UFGroup::getEditHTML(NULL, '', NULL, TRUE, TRUE, NULL, FALSE, $civicrm->getCtype());
 
   $form['civicrm_profile_register'] = array(
     '#markup' => \Drupal\Core\Render\Markup::create($html),


KarinG:
Jenkins re test this please

KarinG:
Yes - this works - now only getting one Contact record in CiviCRM when I add User on the Drupal side;

KarinG:
test this please

dsnopek:
Hm. I just tested this, and now I'm getting no Contact record in CiviCRM when I add a Drupal user. :-/

The steps I followed:

1. First without the patch, add a user called "New User 1" (as an admin on `/admin/people/create`)
2. Go to "Search" -> "Find contacts"
3. I see two contacts for "New User 1", one with the e-mail and one without: selection_147
4. Checkout the code in this PR
5. Create another new user called "New User 2" in the same way
6. Go back to "Search" -> "Find contacts"
7. I see no contact record for "New User 2"

yaazkal:
@dsnopek thanks for testing. I retest it and what I can see is that what you experienced is probably a cache on the search result (refreshing the search page is not enough), if you sort the results by name (or other column) you will be able to see that actually the contact has been created.

I found the _Constituent Summary_ on the Reports section to be more "effective" (refreshing the page is enough to show the new record).

I have tried when an admin adds a user in drupal and also when a visitor register on the drupal site. Both worked for me.

KarinG:
- @dsnopek - I followed your exact same steps and my 3. is identical (two users one with and one without email address);
- Search -> Find Contacts -> Search is fine (should not be cached); 
- I had to apply this one-line patch manually as my .module file on this live instance is somewhat older; 
- I can't seem to get Jenkins to execute the tests on this [will ask];

seamuslee001:
Jenkins test this please

dsnopek:
>  I retest it and what I can see is that what you experienced is probably a cache on the search result (refreshing the search page is not enough), if you sort the results by name (or other column) you will be able to see that actually the contact has been created.

Wow, ok, changing the sort does show the new contacts!

I actually have a patch related to caching that I haven't contributed yet - I wonder if that'll fix it...

In any case, this PR is testing good! I don't really understand the change - I tried walking through the code in `\CRM_Core_BAO_UFGroup::getEditHTML()` and couldn't really figure out what those arguments did.

dsnopek:
I just tested again and my caching change didn't fix it (see https://github.com/civicrm/civicrm-drupal/pull/511) - it must be cached in something else.

KarinG:
Glad that was a caching issue!

seamuslee001:
@dsnopek David, if you had performed the search already it might be cached in the PrevNextCache

see `CRM/Core/BAO/PrevNextCache.php` in civicrm-core. In theory what might be right is that when a contact is added like this we flush the contact caches from core perhaps?

dsnopek:
> see CRM/Core/BAO/PrevNextCache.php in civicrm-core. In theory what might be right is that when a contact is added like this we flush the contact caches from core perhaps?

Maybe? I guess it depends on what the purpose of that cache is. If it's to ensure results don't move around as you're paging through a long list, then it needs to NOT be flushed when a new contact is added. Skimming through the code, I couldn't really determine the high-level "why" behind that cache, so, I'm not really sure...

Probably something for a new issue!

dsnopek:
I just retested this code after the other user registration PR's were merged (see https://github.com/civicrm/civicrm-drupal/pull/513 & https://github.com/civicrm/civicrm-core/pull/11507) and it's still working as expected for me :-)

KarinG:
Awesome - thank you! @totten - this can be merged.

seamuslee001:
@eileenmcnaughton @totten this can be merged. Please note that CI php version does not allow for proper CI testing for D8 hence the x

eileenmcnaughton:
I'm merging this based on @KarinG & @seamuslee001 comments - I haven't looked at it myself but I think their approval is appropriate for merging this

PR #512, opened 2018-01-07
CRM-21628: Allows selection of State given a Country in the Profile creation/edit form (alternate implementation)

dsnopek:
This is an alternate implementation to fix the same bug as https://github.com/civicrm/civicrm-drupal/pull/509

See this comment for the details: https://github.com/civicrm/civicrm-drupal/pull/509#issuecomment-355856161

---

 * [CRM-21628: Can't select country state from the registration form in Drupal 8](https://issues.civicrm.org/jira/browse/CRM-21628)

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -177,6 +177,9 @@ function civicrm_user_view(array &$build, \Drupal\user\UserInterface $account, \
       '#weight' => $conf['weight'],
     );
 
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Looks like a slam dunk PR to merge, this has been tested by a 3rd source, but documented on #509 . Merging #512

PR #513, opened 2018-01-11
CRM-21652: Fix numerous problems with anonymous users registering themselves

dsnopek:
See https://issues.civicrm.org/jira/browse/CRM-21652

Companion PR for civicrm-core: https://github.com/civicrm/civicrm-core/pull/11507

### DIFF
--- a/civicrm.user.inc
+++ b/civicrm.user.inc
@@ -24,6 +24,13 @@ function civicrm_user_insert(AccountInterface $account) {
   }
   \CRM_Core_BAO_UFMatch::updateContactEmail($contact_id, $account->getEmail());
 
+  // If already in CiviCRM, we don't do then next bit because it will cause
+  // the current request to exit, stopping execution.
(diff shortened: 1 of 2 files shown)


civicrm.user.inc
```diff
@@ -24,6 +24,13 @@ function civicrm_user_insert(AccountInterface $account) {
   }
   \CRM_Core_BAO_UFMatch::updateContactEmail($contact_id, $account->getEmail());
 
+  // If already in CiviCRM, we don't do then next bit because it will cause
+  // the current request to exit, stopping execution.
+  $config = CRM_Core_Config::singleton();
+  if ($config->inCiviCRM) {
+    return;
```
dsnopek, line 31: The Drupal 7 module does a similar thing. See https://github.com/civicrm/civicrm-drupal/blob/7.x-master/civicrm_user.inc#L64

PR #515, opened 2018-01-22
CRM-21689 - civicrm rules - cannot select contact from data selector

jitendrapurohit:
* [CRM-21689: civicrm rules - cannot select contact from data selector](https://issues.civicrm.org/jira/browse/CRM-21689)

### DIFF
--- a/modules/civicrm_rules/civicrm_rules_action.inc
+++ b/modules/civicrm_rules/civicrm_rules_action.inc
@@ -63,7 +63,7 @@ The content of the email is:
   $defaults = array(
     'parameter' => array(
       'contact' => array(
-        'type' => 'contact',
+        'type' => 'civicrm_contact',
         'label' => t('Contact'),
         'save' => TRUE,
       ),


jackrabbithanna:
Line 92 also has the same line where type = 'contact'. Should that be changed for that action? Do you happen to have CiviCRM Entity installed in the instance you testing?

In your issue, the event "CiviCRM Membership has been created" I know is provided by CiviCRM Entity, not any Core drupal module. So the Rules provided by CiviCRM Entity use the native Drupal entity types, which for contact the entity type name is, civicrm_entity...

My worry is that there maybe some other Rules events or actions that would depend on the type being contact here, because of CiviCRM Core's non-entity based Rules integration. 

There is a "Contact has been created" Rules event, provided by the Core Civi Rules module. And I it relies on 'contact', and not 'civicrm_contact' in those places.

The contact data used by CiviCRM Core Rules module and the CivICRM Entity contact rules events/actions are incompatible. I had written an CiviCRM Entity version of a add contact to group and remove contact from group actions. Check it out in 2.0

It looks like in your case you should use the "Assign CiviCRM Contact to Group" action in your Rule, and it will work. 

Recommend contributor test my recommendation.

Recommend not merging PR

jitendrapurohit:
`Assign CiviCRM Contact to Group ` action from https://github.com/jackrabbithanna/civicrm_entity works for me. Agree with the above point, closing this PR. Thanks @jackrabbithanna

PR #516, opened 2018-02-05
Update civicrm_handler_field_contact_image.inc

yurg:
Gives "Notice: Undefined index: image_style in civicrm_handler_field_contact_image->render() (line 130 of sites\all\modules\civicrm\drupal\modules\views\civicrm\civicrm_handler_field_contact_image.inc)." if "Custom Files Directory" at civicrm/admin/setting/path?reset=1 doesn't contain "[civicrm.files]" ( in my case default directory has been set as "custom/"  only; changing it to be [civicrm.files]/custom has got Drupal image styles back for CiviCRM Contact Image field in Views )


jackrabbithanna:
This is just adding a informational prefix. Its good to have some help text. Recommend merging this.

jackrabbithanna:
Merging #516

PR #519, opened 2018-02-09
CRM-21758: Add created_id to Views integration

MegaphoneJon:
## Overview

It's not possible to add the field "created_id" to a View displaying items of type CiviCRM Event.  While it's possible with `civicrm_entity`, this won't provide a relationship to Contact entities.  With this patch, you can create views that only show the events created by the logged-in user.

## Before
`created_id` is not accessible to Views.

## After
`created_id` is accessible to Views.

---

 * [CRM-21758: Add event "created_id" to Drupal Views](https://issues.civicrm.org/jira/browse/CRM-21758)

### DIFF
--- a/modules/views/components/civicrm.event.inc
+++ b/modules/views/components/civicrm.event.inc
@@ -343,6 +343,31 @@ function _civicrm_event_data(&$data, $enabled) {
       'handler' => 'views_handler_sort',
     ),
   );
+  // Event created_id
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
I can tell by just reading this code that this is perfect for adding the basic created_id column Views field, filter, sort, and relationship integration. I recommend merging this

jackrabbithanna:
Merging #519

PR #521, opened 2018-02-14
CRM-16541: drush civicrm-sql-dump: always use --routines

com2:
crm-16541 Allow extra options to be passed to mysqldump. Include --routines by default. Same as #520


mlutfy:
jenkins, test this please

mlutfy:
This patch makes sense to me. I tested with/without, and can confirm that with this patch, the civicrm functions are correctly exported.

However:

- This patch is against 4.6. We should fix against 5.x instead.
- Not a blocker, but we should consider d8 too. There is low activity on this queue because D7 is EOL.

(I cancelled the tests when I noticed that this was for 4.6)

seamuslee001:
I'm going to close this as 4.6 is no longer supported and as per #520 the same changes have been included in 7.x-master for ages

PR #523, opened 2018-03-01
CRM-21818: set timezone when initializing civicrm

kerasai:
**JIRA Issue:** https://issues.civicrm.org/jira/browse/CRM-21818

It seems as though the timezone handling was omitted when porting to Drupal 8.

See the [Drupal 7 implementation](https://github.com/civicrm/civicrm-drupal/blob/7.x-master/civicrm.module#L223.).

### DIFF
--- a/src/Civicrm.php
+++ b/src/Civicrm.php
@@ -49,7 +49,7 @@ class Civicrm {
     }
 
     // Initialize the system by creating a config object
-    \CRM_Core_Config::singleton();
+    \CRM_Core_Config::singleton()->userSystem->setMySQLTimeZone();
 
     // Mark CiviCRM as initialized.
     $this->initialized = TRUE;


seamuslee001:
Jenkins ok to test

seamuslee001:
ping @dsnopek @jackrabbithanna

src/Civicrm.php
```diff
@@ -49,7 +49,7 @@ public function initialize() {
     }
 
     // Initialize the system by creating a config object
```
dsnopek, line 51: The comment here should probably be updated

dsnopek:
Otherwise, calling this method is good! It's needed to get correct timezone handling. (FYI, @kerasai is a colleague at MDW ;-))

dsnopek:
We've been using this in production for over a month. Is there anything we can do to help get this merged?

totten:
To facilitate distribution via `composer`, the Civi-D8 code is being migrated to a new repository with a slightly different branch naming convention (ie `civicrm-drupal@8.x-master` becomes `civicrm-drupal-8@master`). The repo is setup is now, and we need to wind-down or migrate any open D8 PRs.

Here's a quick review:

(*CiviCRM Review Template WORD-1.1*)



* ([`r-explain`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-explain)) __Pass__
* ([`r-test`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-test)) __Pass__: The test-failure reported in Github UI can't possible be accurate -- the real issue is that we don't have a Civi-D8 test suite right now. That's not the PR's problem.
* ([`r-code`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-code)) __Pass__
* ([`r-doc`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-doc)) __Pass__
* ([`r-maint`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-maint)) __Pass__: We don't have a Civi-D8 test suite right now. That's not the PR's problem.
* ([`r-run`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-run)) __Pass__: Trusting @dsnopek's comments
* ([`r-user`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-user)) __Pass__: Small change and generally aligned with common practice in other CMS integrations
* ([`r-tech`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-tech)) __Pass__: Small change and generally aligned with common practice in other CMS integrations. Even if there were an `r-tech` issue, it'd be OK to change slightly since we're still pre-release.

PR #528, opened 2018-04-09
civicrm.config.php.drupal: fix logic to find the civicrm.settings.php…

mlutfy:
This PR brings back the logic for finding the `civicrm.settings.php`, including support for multi-site (`sites/example.org/civicrm.settings.php`).

I did a bit of cleaning up, but otherwise kept most of the D7 code, while keeping the D8 code that was already there (ex: throwing an exception: I'm not sure if it's a good idea, but I left it as-is).

I tested this on D8 multi-site.

For REST API users, this has a companion PR https://github.com/civicrm/civicrm-core/pull/11958.

### DIFF
--- a/civicrm.config.php.drupal
+++ b/civicrm.config.php.drupal
@@ -32,33 +32,95 @@
  * 10. $confdir/default
  *
  */
-
 function civicrm_conf_init() {
-    global $skipConfigError;
+  global $skipConfigError;
 
-    static $conf = '';
+  static $conf = '';
 
-    if ($conf) {
-        return $conf;
+  if ($conf) {
+    return $conf;
+  }
+
+  /**
+   * We are within the civicrm module, the drupal root is 2 links
+   * above us, so use that
+   */
+  $currentDir = dirname(__FILE__) . DIRECTORY_SEPARATOR;
+  if (file_exists($currentDir . 'settings_location.php')) {
+    include $currentDir . 'settings_location.php';
+  }
+
+  if (defined('CIVICRM_CONFDIR') && !isset($confdir)) {
+    $confdir = CIVICRM_CONFDIR;
+  }
+  else {
+    // make it relative to civicrm.config.php, else php makes it relative
+    // to the script that invokes it
+    $moduleDir  = 'sites' . DIRECTORY_SEPARATOR . 'all' . DIRECTORY_SEPARATOR . 'modules';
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

Does this code work both when the vendor directory is in the webroot and when it's not? I've been trying to read through it to determine that, but the code is quite complex. I'll try to find time to test it later, but this week is Drupalcon so I might not get to it. :-)

mlutfy:
Indeed, I had difficulty understanding the code, but it tries to support:

- Drupal site/vhost aliases (by looking in `sites.php`)
- Drupal support for, say, "abc.www.example.org" -> `sites/example.org` (while searching for the correct precedence, i.e. `sites/abc.www.example.org` then `sites/www.example.org`, etc).

I'm not sure about the `vendor` directory is outside the webroot. Does the code currently support it? (I feel like we would need to add an entry in the `$candidates`). Re-reading the code, maybe the check in line 74 should be moved to the end, or simply removed.

i.e. this part:
```
if (!file_exists($confdir) && !$skipConfigError) {
  echo "Could not find valid configuration dir, best guess: $confdir<br/><br/>\n";
  exit();
}
```

dsnopek:
> I'm not sure about the vendor directory is outside the webroot. Does the code currently support it?

Yes.

Just to be clear, this means you have a directory structure like:

- web/ <-- where Drupal is
- web/sites/SITE/settings.php
- vendor/
- vendor/civicrm/civicrm-core

... where 'web' is frequently used, but sometimes other directory names are used.

In any case, I'm not sure if your changes affect this at all, since I haven't tested this.

jackrabbithanna:
Great, we've had some problems in 4.6.29 with Stripe IPNs and it ended up being that civicrm_settings.php could not be found. We use a multi-site...well test this...

Also if this is needed for D8, that is supporting vendor/civicrm/civicrm-core it would be great to get it in. 

@dsnopek  definitely knows what it needs.

dsnopek:
I tested this and worked on our platform with the vendor directory outside the web root! I walked through the code in the line debugger to see what it was doing, and the bit where it's loading `CIVICRM_CONFDIR` from 'settings_location.php' is what's giving it the right path for us.

So, :+1:!

mlutfy:
jenkins, test this please

mlutfy:
Closing, since D8 has moved repos.

PR #531, opened 2018-06-05
dev/core/#152 - AdvMulti-Select cleanup

colemanw:
A little more cleanup following on from https://github.com/civicrm/civicrm-core/pull/12238

@jitendrapurohit can you look at this too?


jitendrapurohit (approved):
Tested on views UI by navigating to some pages and adding custom fields. Works fine.

PR #533, opened 2018-06-20
Require the 5.2 branch of CiviCRM

borisson:


kewljuice:
I can confirm that this patch is needed for Drupal 8 to work with CiviCRM 5.2 via composer

totten:
@borisson @kewljuice - agree this part of `composer.json` should be updated. However, to provide better support for `composer`, development for the Civi-D8 integration was migrated in April to https://github.com/civicrm/civicrm-drupal-8/. (In packagist, that's now `civicrm/civicrm-drupal-8`.) This is because `composer` does not work well with branches named `8.x-{VERSION}`.

There is a similar PR in `civicrm-drupal-8` -- https://github.com/civicrm/civicrm-drupal-8/pull/2. Could you try that?

borisson:
We'll update that PR for more discussion. Closing this one.

PR #534, opened 2018-07-15
Drupal Views CiviCRM: Allow Participant record to be used from Event.

JohnFF:
Under Drupal Views, there's no relationship that allows a Contact to be connected to an Event by a Participant record.

Outstanding feature: currently using this allows all the default Contact fields, but not custom ones.

### DIFF
--- a/modules/views/civicrm/civicrm_handler_relationship_participant.inc
+++ b/modules/views/civicrm/civicrm_handler_relationship_participant.inc
@@ -1,0 +1,71 @@
+<?php
+
+/**
+ * This relationship handler is used when joining the civicrm_relationship table
+ * to the civicrm_contact table. This handler allows us to optionally add conditions
+ * to the join clause based on relationship_type_id and is_active.
+ */
+class civicrm_handler_relationship_participant extends views_handler_relationship {
+  static $participant_types;
+
+  /**
+   * Preload the list of relationship_types and store in the static variable
+   * $relationship_types
+   */
+  public function construct() {
+    parent::construct();
+
+    if (!civicrm_initialize()) {
+      return;
+    }
+    require_once 'CRM/Core/PseudoConstant.php';
+
+    self::$participant_types = CRM_Core_PseudoConstant::participantRoles();
+  }
+
+  /**
+   * Add additional options for relationship_type and relationship_state
+   * to the view. By defining these here, Views will take care of saving the
+   * values submitted from the options form.
+   */
+  public function option_definition() {
+    $options = parent::option_definition();
+    $options['participant_type'] = array('default' => 0);
+    return $options;
+  }
+
+  /**
+   * Add relationship_type drowndown and relationship_state checkbox to
+   * relationship configuration form.
+   */
+  public function options_form(&$form, &$form_state) {
+    parent::options_form($form, $form_state);
+
+    $form['participant_type'] = array(
+      '#type' => 'select',
+      '#multiple' => TRUE,
+      '#title' => 'Choose a specific Participant type',
+      '#options' => self::$participant_types,
+      '#description' => t('Choose to limit this relationship to one or more specific types of CiviCRM Participant.'),
+      '#default_value' => $this->options['participant_type'],
+    );
+  }
+
+  /**
+   * Modify the default views relationship query to optionally specify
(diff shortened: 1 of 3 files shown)


JohnFF:
@eileenmcnaughton this is what I meant by code that allows Drupal Views to have a relationship between Events and Contacts via Participant record. The default Contact fields are usable via the relationship, but custom ones are not. Of course, you can create a new relationship joining the Contact ID to the Contact Record which allows custom fields, but this seems extra - it'd be good to have them visible automatically if possible.

JohnFF:
Hi @colemanw - do you know when this PR may make its way into a release? Also, the check above sounds borked - 3 spaces? Surely it should be 2. First bit of feedback you ever gave me IIRC... :)

modules/views/civicrm/civicrm_handler_relationship_participant.inc
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Add additional options for relationship_type and relationship_state
+   * to the view. By defining these here, Views will take care of saving the
+   * values submitted from the options form.
+   */
+   public function option_definition() {
+    $options = parent::option_definition();
+    $options['participant_type'] = array('default' => 0);
+    return $options;
+  }
```
jackrabbithanna: add space here

jackrabbithanna:
Some small code style test failures

check out the spaces around lines 31-35...
32-35 need an additional space before the code

Fix those up and we'll rerun the Jenkins tests to see if that gets them

colemanw:
Code looks good to me aside from the nitpicks from Jenkins :)
Has anyone tested it out by running it? Soon as we get someone saying "works for me" then I think we're good.

modules/views/civicrm/civicrm_handler_relationship_participant.inc
jackrabbithanna: actually do 26-34 have one too many space

JohnFF:
Hi @colemanw @jackrabbithanna ,
I'm using this myself - but I have a question... This relationship allows access to default Contact fields. Fine. But in order to use Custom Contact Fields you have to create a new relationship to contact. 

I think a big improvement would be to allow contact custom fields across the Participant relationship, seeing as the default behaviour is to allow standard contact fields. How can we do that?

colemanw:
I'm not familiar enough with the Views-CiviCRM code, sorry.

JohnFF:
@torrance @seamuslee001 @laryn < I'm asking you because you have recent CiviCRM Drupal Views Relationships commits in a similar area. Do you know the answer to the above question? How a Relationship can access not just the standard fields on a Contact but the Custom fields as well?

laryn:
Sorry, I don't know – but I do agree it would be an improvement!

JohnFF:
Requires https://github.com/civicrm/civicrm-core/pull/12622 to be merged before this can be!

colemanw:
I believe it no longer needs the above?

modules/views/civicrm/civicrm_handler_relationship_participant.inc
```diff
@@ -0,0 +1,71 @@
...
+   * Preload the list of relationship_types and store in the static variable
+   * $relationship_types
+   */
+  public function construct() {
+    parent::construct();
+
+    if (!civicrm_initialize()) {
+      return;
+    }
+    require_once 'CRM/Core/PseudoConstant.php';
```
seamuslee001, line 21: is this needed if civicrm is initialised?

eileenmcnaughton: nope - but I discussed with @JohnFF & he's gonna use the api instead

JohnFF:
True.

modules/views/civicrm/civicrm_handler_relationship_participant.inc
JohnFF: @seamuslee001 the CiviCRM initialise check is still needed most likely, because the API call comes next. I am now using the API, so the PseudConstant.php require is removed.

JohnFF:
@eileenmcnaughton @colemanw @seamuslee001  - new PR made. This one is closed.

PR #536, opened 2018-07-22
Update civicrm_rules_event.inc

yurg:
There are 3 Rule Events available for CiviCRM Contact: 

* After updating an existing civicrm contact
* CiviCRM Contact has been updated
* Contact has been updated

A little bit hard to find what is the difference. Let us help users by telling that "Contact has been updated" comes from Event.
Perhaps, would be even better to move this one under CiviCRM Events option group.

### DIFF
--- a/modules/civicrm_rules/civicrm_rules_event.inc
+++ b/modules/civicrm_rules/civicrm_rules_event.inc
@@ -47,7 +47,7 @@ function civicrm_rules_get_event() {
     ),
     'contact_edit' =>
     array(
-      'label' => t('Contact has been updated'),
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
'CiviCRM Contact has been updated' is coming from the CiviCRM Entity module...

yurg:
@jackrabbithanna  Thank you for note! The only mystery to resolve is, where 'After updating an existing civicrm contact' comes from (guess is form CiviCRM-Drupal too though ) and how is it differs from  'CiviCRM Contact has been updated' and why do we need both.

PR #538, opened 2018-07-26
dev/drupal/27 Custom fields with duplicate labels are not displayed correctly in views

JKingsnorth:
Steps to recreate:

- Create a custom field group associated with organisations, add a field with the label 'Role' (eg: ID 1)
- Create a custom field group associated with individuals, add a field with the label 'Role' (eg: ID 2)
- Create a view which exposes the second 'role' field on a contact record
- The field is always empty

Reason:

- The custom field handler is searching for the field ID based on the label of the field, which is a loose search, it returns the first match, which is ID 1 in the example above.

Remedy:

- Tighten the custom field handler to search for the ID of the field based on the unique column name.

### DIFF
--- a/modules/views/civicrm/civicrm_handler_field_custom.inc
+++ b/modules/views/civicrm/civicrm_handler_field_custom.inc
@@ -70,7 +70,7 @@ class civicrm_handler_field_custom extends views_handler_field {
         if (!is_null($value)) {
           // get the field id from the db
           if (!empty($this->definition['title'])) {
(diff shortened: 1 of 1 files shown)


JKingsnorth:
I've just noticed that there is no index on the 'column_name' field, which may reduce performance slightly on websites with a large number of custom fields? Happy to take advice on whether this is an issue worth following up; but I expect that the performance impact will be too small to easily test? As a sidenote - the 'column_name' field seems like one that should have an index on it.

An alternative approach to this PR would be to extract the field ID from the alias. Eg: extract the '6' from 'role_6'.

PR #539, opened 2018-08-05
Drupal Views CiviCRM: Allow Participant record to be used from Event.

JohnFF:


JohnFF:
@colemanw @eileenmcnaughton @seamuslee001 @jackrabbithanna < this is the new PR :) it has the changes suggested.

Please note that the civicrm_api check is still needed for the API call.

If someone else could please test this, and let Coleman know if you're happy for it to be merged, that would be amazeballs, and I'd happily return favour in future.

modules/views/civicrm/civicrm_handler_relationship_participant.inc
```diff
@@ -0,0 +1,74 @@
+<?php
+
+/**
+ * This relationship handler is used when joining the civicrm_relationship table
```
seamuslee001: @JohnFF are you able to update this code comment

modules/views/components/civicrm.core.inc
```diff
@@ -157,6 +157,18 @@ function _civicrm_core_data(&$data, $enabled) {
...
 
+  $data['civicrm_event']['civicrm_participant'] = array(
+    'real field' => 'id',
+    'title' => t('CiviCRM Participant (starting from Event)'),
+    'help' => t('Allows a Participant Contact relationship from an Event'),
+    'relationship' => array(
+      'base' => 'civicrm_participant',
+      'base field' => 'event_id',
+      'handler' => 'civicrm_handler_relationship_participant',
+      'label' => t('CiviCRM Participant (starting from Event)'),
```
colemanw: I don't understand this string (note that it appears twice, if we decide to change it).
What about "CiviCRM Event Participant"

JohnFF: Hi @colemanw - this string appears twice in the same way that it does for CiviCRM Relationship. I also looked to "starting from Contact A" as guidance, although here I see it's probably not necessary. I've adjusted the PR accordingly.

JohnFF:
image
image

JohnFF:
You can check out the end results here: http://civicrm-drupal-john886170.codeanyapp.com (site may time out, I'm unsure atm).

colemanw:
* ([`r-explain`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-explain)) __Pass__
* ([`r-test`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-test)) __Pass__
* ([`r-code`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-code)) __Pass__
* ([`r-doc`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-doc)) __Pass__
* ([`r-maint`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-maint)) __Pass__
* ([`r-run`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-run)) __Pass__: Tested on demo site
* ([`r-user`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-user)) __Pass__: Meets user expectations for Views joins
* ([`r-tech`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-tech)) __Pass__: consistent with other code.

PR #540, opened 2018-08-17
Don't render an empty mailto link for empty email addresses

JKingsnorth:
If you add an 'email address' field to a views display, and choose to output it as a mailto link then an empty link is created when the email address is empty; with the link of just 'mailto:'.

This patch makes a change to ensure that an email address is present before outputting a link around it.


seamuslee001:
This looks sensible and fine merging

PR #541, opened 2018-08-31
PHP 7.1 incompatibility error fix for -> Error: Using $this when not in object context in civicrm_form_data()

tmannell:
I was getting this error when attempting to edit civi profiles exposed to drupal user profiles.

 Error: Using $this when not in object context in civicrm_form_data() (line 590 of sites/all/modules/civicrm/drupal/civicrm.module).

Changing the $this to the available $edit var, which appears to contain the form, seemed to solve the problem.


jitendrapurohit (approved):
I was able to replicate this on one of our sites and this PR seems to have fixed it for us. Confirming that it is a good/correct change and should be merged 👍

eileenmcnaughton:
merging per @jitendrapurohit review - looks like it was checked a while back but no-one spotted that

PR #542, opened 2018-09-11
dev/core#381 - Clear session vars while retrieving the files with id a…

jitendrapurohit:
…nd eid

Details in https://lab.civicrm.org/dev/core/issues/381#note_8535


mlutfy:
From the Gitlab issue, it's not clear if someone tested this patch?

colemanw:
I see that this has been reviewed and tested in GitLab.

PR #544, opened 2018-10-04
Bug fix in civicrm.module

albionbrown:
Replaced `$this` with `$edit` in `civicrm_form_data()` when not in object context. 
Ignore the README I made

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,0 @@
-This fork fixes a bug in the civicrm.module file around lines 587-590, regarding the error message `Error: Using $this when not in object context in civicrm_form_data() (line 591 of /path/to/site/sites/all/modules/contrib/civicrm/drupal/civicrm.module).` 
--- a/README.md
+++ b/README.md
@@ -1,0 +1,1 @@
+This fork fixes a bug in the civicrm.module file around lines 587-590, regarding the error message `Error: Using $this when not in object context in civicrm_form_data() (line 591 of /path/to/site/sites/all/modules/contrib/civicrm/drupal/civicrm.module).` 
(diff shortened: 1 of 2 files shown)


MegaphoneJon:
Many thanks for the contribution Josh!  Are you able to resubmit (perhaps with a forced rebase) without the README so it's mergeable?  You can do it now or after review takes place.  If you have questions on how to do it, feel free to ask here, or in the ~dev-newcomers channel on https://chat.civicrm.org.  Thanks, and welcome!

albionbrown:
Hi Jon!

I forgot to reply to you yesterday, check the latest commit I think it's
done now (never done a rebase, but I think it's right!).

Josh

MegaphoneJon:
This isn't a rebase, but that's OK - the committer can also deal with it on their end.

For whatever reason it's a little harder to get folks who know this repo to review patches compared to the "civicrm-core" repo, and I'm unfortunately not really qualified here.  I'll work on finding someone.  Thanks again for taking the time.

jackrabbithanna:
How did the problem manifest before this patch? Is there a process that we can follow to replicate the issue and confirm the fix?

jackrabbithanna:
This is a duplicate of https://github.com/civicrm/civicrm-drupal/pull/541/

albionbrown:
Oh sorry I forgot to note how to reproduce! 
 
1. Log into a Drupal user who can administer CiviCRM and Drupal users
2. Find a contact in Civi whose account is not 'linked' to the Drupal account (The user's account has never been logged into and therefore has no record in civicrm_uf_match table to map Civi ID to Drupal User ID)
3. Go to that contact's associated Drupal user page
4. Click on 'Edit'
5. Finally click on 'User Profile'

petednz:
i am seeing this same issue on some contacts where there is a uf_match - and where the Name and Address profile on the /user page does show the civi data - but clicking 'edit Name and Address' throws the error. noting this was a WP site that we moved to Drupal, and cleared out all unwanted uf_match rows and fixed up the remaining ones to work. we are seeing this error on newly created users though as well

mlutfy:
@petednz Can you apply the patch and confirm whether it fixes the issue?

I would still recommend merging #541 instead, because the commit log will be cleaner.

eileenmcnaughton:
#541 merged - closing this as it seems superceded

PR #546, opened 2018-10-14
dev/drupal#33 Implement userFrameworkFrontend configruation object fo…

seamuslee001:
…r Drupal

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -214,6 +214,17 @@ function civicrm_initialize() {
     // initialize the system by creating a config object
     $config = CRM_Core_Config::singleton();
 
+    // Get the menu items.
+    $args = explode('?', $_GET['q']);
+    $path = $args[0];
+
+    // Get the menu for above URL.
+    $item = CRM_Core_Menu::get($path);
+
+    if (!empty(CRM_Utils_Array::value('is_public', $item))) {
+      $config->userFrameworkFrontend = TRUE;
+    }
+
     // Add module-specific header elements
     $header = civicrm_html_head();
     if (!empty($header)) {


civicrm.module
```diff
@@ -214,6 +214,17 @@ function civicrm_initialize() {
...
 
+    // Get the menu items.
+    $args = explode('?', $_GET['q']);
+    $path = $args[0];
+
+    // Get the menu for above URL.
+    $item = CRM_Core_Menu::get($path);
+
+    if (!empty(CRM_Utils_Array::value('is_public', $item))) {
+      $config->userFrameworkFrontend = TRUE;
```
totten, line 225: The good and the bad:

* The really good thing is that it makes Civi's internal metadata (`$config->userFrameworkFrontend`) more consistent across the various CMS's -- i.e. the flag gets a similar meaning on D7, WP, J, etal. It also allow some cleanup re: https://github.com/civicrm/civicrm-core/pull/12929 so that the theming options are similar in each environment.
* From an `r-tech` perspective, the tricky thing is that there is other functionality (`rgrep userFrameworkFrontend`) which keys off of this variable - and so one needs to give a think to whether D7 folks hitting those linse will perceive that the change as regressive.

mattwire: It would be really nice to get this consistent with other CMSs.  @totten I assume this has been discussed but what does is_public actually mean and how is it set? In other words, is it likely to be reliable?

seamuslee001: @mattwire is_public is set in the xml/menu routes and i this is generally code for this route is publicly accessible needed or something (actually realised it may have some perms attached)

see https://github.com/civicrm/civicrm-core/blob/265ce2a90a624618bdef2f3a784956841de96335/CRM/Core/DAO/Menu.php#L398  https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/Invoke.php#L244  https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/Invoke.php#L254

eileenmcnaughton: @seamuslee001 I would note that we discussed one method we sometimes use to grandfather things in is permitting 'define' - I see this as a setting being something we're expecting to keep as an option whereas a define being something we support to allow the advance guard to test . a change we think we want to make generically but want more real world validation of.

totten:
jenkins, test this please

AkA84:
hello, coming from https://lab.civicrm.org/dev/core/issues/378#note_20974 :) was wondering what is the status of this PR? I guess best case scenario this will be out in 5.17?

eileenmcnaughton:
@AkA84 I am OK with targeting 5.16 to be in the same version as the main theme fix if we can agree the right fix - I *think* this addresses the concerns raised here https://github.com/civicrm/civicrm-core/pull/14876 - although the setting might still be exposed in the UI for drupal  - I didn't check. Do you want to test it & see if you can make it work in your context.

eileenmcnaughton:
Closing as replaced by https://github.com/civicrm/civicrm-core/pull/14876

PR #547, opened 2018-10-26
civicrm_handler_field_link_pcp: avoid PHP notice

mlutfy:
This function generated PHP notices 'Undefined property: stdClass::$unknown',
yet the field_alias is not used anyway in the render_link() function. The change
proposed here is consistent with similar changes elsewhere (ex: 8da894b5).

My view was fairly typical: contact names and page URL (we did rewrite the page URL field so that it links on the page title). Full view: http://paste.debian.net/hidden/5d54a877/


jackrabbithanna:
This is a good fix, I recommend merging.

seamuslee001:
Merging as per Mark's recommendation

PR #550, opened 2018-11-16
Add a Views Relationship handler for relating a Contact to their Contributions.

anemirovsky:
The existing Views integration allows for pulling in Contributions related to
a Contact via indirect methods, like from a related Participant or Membership
record. Doing so does not allow Contribution Custom Fields to be used in the
View, though. This change adds an explicit Relationship between Contacts and
Contribution which, when added to a View, allows for pulling in Custom Field
data for Contributions.


stesi561:
I'm pretty sure this is d7 specific and therefore should now be closed.

PR #552, opened 2018-12-11
Fix path for civicrm.settings.php when installed in profiles/

sluc23:
When CiviCRM is installed in a Drupal's distribution profile, under **profiles/<profile_name>/modules/[contrib]**, the funciton `civicrm_conf_init()` doesn't return the proper path for file **civicrm.settings.php**.

In most of cases CiviCRM works fine without this patch, but it fails i.e. when calling an API using REST, returns this error:
```
Could not load the settings file at: /var/www/drupal/profiles/myCustomProfile/modules/contrib/civicrm/../..//default/civicrm.settings.php
```


seamuslee001:
Jenkins add to whitelist

sluc23:
Can one of the admins verify this patch?

mlutfy:
jenkins, test this please

PR #553, opened 2018-12-13
(NFC) Update Copyrite header to change to 2019

seamuslee001:
ping @totten @eileenmcnaughton


seamuslee001:
Merging as per the tag

PR #554, opened 2018-12-13
Update version to match core and update copyright header to be 2019

seamuslee001:
ping @eileenmcnaughton @totten


seamuslee001:
Merging as per the tag

PR #560, opened 2019-02-15
fix checksum validation in Drupal

mlutfy:
(backport)

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -584,10 +584,10 @@ function civicrm_form_data($edit, &$user, $category, $reset, $doNotProcess = FAL
     // do not allow edit for anon users in joomla frontend, CRM-4668, unless u have checksum CRM-5228
     $config = CRM_Core_Config::singleton();
     if ($config->userFrameworkFrontend) {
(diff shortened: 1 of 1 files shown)


seamuslee001:
@eileenmcnaughton thoughts?

eileenmcnaughton:
Wasn't it just a coding error not a php 7.1 error - ie. no in object context, don't use $this

I'm sure you know I advocate we run CI on php 7.2 which may have helped

mlutfy:
yep, just a programming error. The subject was from the commit I cherry-picked from master.

JoeMurray:
Updated the title of the PR

mlutfy:
Any objections to merging?

eileenmcnaughton:
@mlutfy I merged this & then realised it's against the wrong branch - it should have gone into the rc & any backport to 5.10 would have been discussed on product-maintenance channel after that - anyway I was able to create cleanly against the rc from your branch https://github.com/civicrm/civicrm-drupal/pull/561 - so fixing there

@seamuslee001 @totten  I think it's fine that this hits the security release - I guess it SHOULD have been in yesterday's release but given we missed that & we are so close to next week's release that will prob be OK

PR #565, opened 2019-03-06
7.x 5.11

seamuslee001:
ping @eileenmcnaughton @totten note that this is in 5.10 but somehow hasn't been brought up to the RC


eileenmcnaughton:
ouch that was a botch up. I'm OK with the backdrop one being merged to - we can 'leave it for a bit' but tomorrow is the latest....

PR #567, opened 2019-04-02
Bump php version to match core requirements

colemanw:
We no longer support php < 5.6.


seamuslee001:
This makes sense to me merging

PR #568, opened 2019-04-02
Fix version number

colemanw:
It appears we've been publishing tarballs with incorrect version numbers since 5.0.

Ping @totten


colemanw:
@totten I think we'll need to merge this and then either manually update this number every time or else teach the releaser script to do it.

seamuslee001:
@colemanw doesn't this cover it? https://github.com/civicrm/civicrm-core/blob/master/distmaker/dists/common.sh#L121

colemanw:
Ah. So it does. Just verified the latest tarball does indeed show the correct version.

PR #569, opened 2019-04-02
Expose "created date" to views

laryn:
This allows date formatting options on the `created date` in views.

### DIFF
--- a/modules/views/components/civicrm.core.inc
+++ b/modules/views/components/civicrm.core.inc
@@ -721,6 +721,22 @@ function _civicrm_core_data(&$data, $enabled) {
       'handler' => 'civicrm_handler_sort_date',
     ),
   );
+  // Created date
+  $data['civicrm_contact']['created_date'] = array(
+    'title' => t('Created date'),
+    'help' => t("The date the contact's record was created."),
+    'field' => array(
+      'handler' => 'civicrm_handler_field_datetime',
+      'click sortable' => TRUE,
+    ),
+    'filter' => array(
+      'handler' => 'civicrm_handler_filter_datetime',
+      'is date' => TRUE,
+    ),
+    'sort' => array(
+      'handler' => 'civicrm_handler_sort_date',
+    ),
+  );
   // Image URL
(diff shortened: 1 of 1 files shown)


modules/views/components/civicrm.core.inc
```diff
@@ -721,6 +721,22 @@ function _civicrm_core_data(&$data, $enabled) {
...
+  $data['civicrm_contact']['created_date'] = array(
+    'title' => t('Created date'),
+    'help' => t("The date the contact's record was created."),
+    'field' => array(
+      'handler' => 'civicrm_handler_field_datetime',
+      'click sortable' => TRUE,
+    ),
+    'filter' => array(
+      'handler' => 'civicrm_handler_filter_datetime',
+      'is date' => TRUE,
```
seamuslee001, line 734: @laryn just a thought should we allow for NULL and NOT NULL filters here as well?

laryn: @seamuslee001 I'm definitely open to that. I simply copied the code from the `modified_date` field in the same area:
https://github.com/civicrm/civicrm-drupal/blob/7.x-master/modules/views/components/civicrm.core.inc#L708-L723

Should NULL and NOT NULL be added to that one as well?

laryn: Further thought: Is there ever a case where a created or modified date would be NULL?

seamuslee001:
Jenkins retest this please

seamuslee001:
Jenkins re test this please

seamuslee001:
Jenkins re test this please

modules/views/components/civicrm.core.inc
yashodha: @laryn no, it shouldn't be the case, the created /modified date would have value always. Modified date = created date if not modified/updated.

yashodha:
@laryn @seamuslee001 looks straight forward, this can be merged as well. Thanks!

yashodha:
@laryn merging this PR

PR #573, opened 2019-04-11
Enable contribution views to filter on empty Receipt and Thank You dates

aydun:
Adding `'allow empty' => TRUE` to filter definitions for dates enables the `is empty` / `is not empty` filter options.

Useful if you want a view showing contributions that have/have not been thanked.

Before: A view filter on Thank You date or Receipt Date does not offer empty/not empty

After: A view filter on Thank You date or Receipt Date does offer empty/not empty


seamuslee001:
This makes sense to me and seems fine code wise

seamuslee001:
I have tested this locally and confirms it works merging

aydun:
Thanks @seamuslee001 !

PR #575, opened 2019-04-11
(NFC) Get phpcs to ignore the use of eval in this file as it is required

seamuslee001:


seamuslee001:
Merging as per the tag

PR #580, opened 2019-05-28
Fixes issue where adding/removing a contact from CiviCRM Group works …

VangelisP:
…with deprecated method on removal

GitLab: https://lab.civicrm.org/dev/drupal/issues/65


eileenmcnaughton:
@jitendrapurohit any chance you can check this

eileenmcnaughton:
This is  obviously it's not going to get more thorough review but I've worked through the discussion and it makes sense so I'm going to merge it

PR #583, opened 2019-08-20
(dev/drupal#79) Fail more gracefully when upgrading on PHP 5.x

totten:
This version requirement officially went up to PHP 7.0 circa Civi 5.14. However, at that time, the upgrade metadata was kept at PHP 5.6 to allow somewhat softer landing for stragglers.  That's no longer possible in Civi 5.16+,

This just gives a clearer error when someone tries to upgrade with PHP 5.x.

Depends: https://github.com/civicrm/civicrm-core/pull/15082

Before
------

When upgrading via drush or Drupal web UI on PHP 5.6, the Civi class-loader fails to initialize.

```
Parse error: syntax error, unexpected ':', expecting '{' 
in /Users/totten/bknix/build/dmaster/web/sites/all/modules/civicrm/vendor/league/csv/src/functions.php on line 33
```

(Approximate call-path: `civicrm.drush.inc | civicrm.module` =>
`civicrm.settings.php` => `CRM_Core_ClassLoader` => `vendor/autoload.php` =>
`vendor/league/csv/src/functions.php`)

After
-----

When upgrading via drush on PHP 5.6, the error points to the actual problem.

```
[bknix-old:~/bknix/build/dmaster/web/sites/all/modules/civicrm] drush civicrm-upgrade-db
CiviCRM requires PHP 7.0.0+. Drush is running PHP 5.6.38.                                                                                                                [error]
```

When upgrading via web UI on PHP 5.6, the error message is similar:

```
CiviCRM requires PHP 7.0.0+. The web server is running PHP 5.6.38.
```

(Note: I tweaked the text to emphasize that the PHP version is determined by the "drush" or "web server" environment - in some systems, these are different environments with different PHP versions.  A phrase like "your version" can be confusing/misleading in those cases.)

Comments
--------

The canonical representation of the minimum PHP version is in `$civicrm_root/CRM/Upgrade/Form.php`.  However, correctly reading that metadata requires loading `civicrm.settings.php`, which triggers the crash.

To work around this, we reproduce the constant and use a unit-test to ensure its continued accuracy.

Also, it seems useful to put the same metadata in `civicrm.info`.  I'm not sure if D7 uses this in a meaningful way, but it's good to be accurate.

### DIFF
--- a/civicrm.info
+++ b/civicrm.info
@@ -4,7 +4,7 @@ version = 7.x-4.7
 package = CiviCRM
 core = 7.x
 project = civicrm
-php = 5.6
+php = 7.0
 
 files[] = civicrm.module
 files[] = civicrm.install
--- a/civicrm.module
+++ b/civicrm.module
@@ -33,6 +33,19 @@ require_once 'civicrm_user.inc';
 define('CIVICRM_UF_HEAD', TRUE);
 
 /**
+ * Minimum required PHP
+ *
+ * Note: This duplicates CRM_Upgrade_Form::MINIMUM_PHP_VERSION. The
+ * duplication helps avoid a dependency-loop. (Reading `Form::MINIMUM_PHP_VERSION`
+ * requires loading `civicrm.settings.php`, but that triggers a parse-error
+ * on PHP 5.x.)
+ *
+ * @see CRM_Upgrade_Form::MINIMUM_PHP_VERSION
+ * @see CiviDrupal\PhpVersionTest::testConstantMatch()
+ */
+define('CIVICRM_DRUPAL_PHP_MINIMUM', '7.0.0');
+
+/**
  * Adds CiviCRM CSS and JS resources into the header.
  */
 function civicrm_html_head() {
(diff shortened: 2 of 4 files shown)


demeritcowboy:
Thanks @totten.  The wording change about "web/drush" vs. "your" is great an on-point, but I wonder if that would actually inform people who **don't know** that you can possibly have two different ones (i.e. they've installed drupal first and it's telling them it's php 7 via web status report, then they go use drush). In the web server case it's fine, but in the drush case, it might be useful to append instructions like "It is possible your command line is configured to use a different php version than the web server. Check your hosting control panel settings to confirm you are using php 7 in all configurations."

seamuslee001:
@totten test failures relate

demeritcowboy:
Off topic but I got pinged by it so of course I looked (personality flaw). Curious that tests are running phpunit4. Expected?

seamuslee001:
Possibly a job config not updated maybe?

totten:
jenkins, test this please

totten:
Good catch on phpunit4. Filed https://github.com/civicrm/civicrm-buildkit/pull/467 and https://lab.civicrm.org/dev/core/issues/1194

totten:
Regarding a longer message - I'm inclined to abstain on that one and invite you to do a PR with better text. (The current draft makes it slightly better than before - your points could make it better-er.)

Aesthetically, I'm less keen on that style of message (e.g. messages about externalities don't age as well)... but, pragmatically, you're quite right in identifying a segment of the audience for which that's a good time for that discussion. Maybe it'd make sense to output a link to sysadmin docs which give the bigger picture about system requirements? (Presumably, the docs would be maintained more sensitively than this message would?)

demeritcowboy:
Ok. I can test this.

totten:
On Mattermost, Dave reported successful testing of the messages for upgrade use-cases. He also identified an issue with the install experience, for which I submitted a separate revision - https://github.com/civicrm/civicrm-core/pull/15089

PR #584, opened 2019-08-21
(dev/drupal#79) Fail more gracefully when upgrading on PHP 5.x

totten:
This version requirement officially went up to PHP 7.0 circa Civi 5.14.
However, at that time, the upgrade metadata was kept at PHP 5.6 to allow
somewhat softer landing for stragglers.  That's no longer possible in Civi
5.16+,

This just gives a clearer error when someone tries to upgrade with PHP 5.x.

Before
------

When upgrading via drush or Drupal web UI on PHP 5.6, the Civi class-loader fails to initialize.

```
Parse error: syntax error, unexpected ':', expecting '{' in /Users/totten/bknix/build/dmaster/web/sites/all/modules/civicrm/vendor/league/csv/src/functions.php on line 33
```

(Approximate call-path: `civicrm.drush.inc | civicrm.module` =>
`civicrm.settings.php` => `CRM_Core_ClassLoader` => `vendor/autoload.php` =>
`vendor/league/csv/src/functions.php`)

After
-----

When upgrading via drush on PHP 5.6, the error points to the actual problem.

```
[bknix-old:~/bknix/build/dmaster/web/sites/all/modules/civicrm] drush civicrm-upgrade-db
CiviCRM requires PHP 7.0.0+. Drush is running PHP 5.6.38.                                                                                                                [error]
```

When upgrading via web UI on PHP 5.6, the error message is similar:

```
CiviCRM requires PHP 7.0.0+. The web server is running PHP 5.6.38.
```

(Note: I tweaked the text to emphasize that the PHP version is determined by
the "drush" or "web server" environment - in some systems, these are
different environments with different PHP versions.  A phrase like "your
version" can be confusing/misleading in those cases.)

Comments
--------

The canonical representation of the minimum PHP version is in
`$civicrm_root/CRM/Upgrade/Form.php`.  However, correctly reading that
metadata requires loading `civicrm.settings.php`, which triggers the crash.

To work around this, we reproduce the constant and use a unit-test to ensure
its continued accuracy.

Also, it seems useful to put the same metadata in `civicrm.info`.  I'm not
sure if D7 uses this in a meaningful way, but it's good to be accurate.

### DIFF
--- a/civicrm.info
+++ b/civicrm.info
@@ -4,7 +4,7 @@ version = 7.x-4.7
 package = CiviCRM
 core = 7.x
 project = civicrm
-php = 5.6
+php = 7.0
 
 files[] = civicrm.module
 files[] = civicrm.install
--- a/civicrm.module
+++ b/civicrm.module
@@ -33,6 +33,19 @@ require_once 'civicrm_user.inc';
 define('CIVICRM_UF_HEAD', TRUE);
 
 /**
+ * Minimum required PHP
+ *
+ * Note: This duplicates CRM_Upgrade_Form::MINIMUM_PHP_VERSION. The
+ * duplication helps avoid a dependency-loop. (Reading `Form::MINIMUM_PHP_VERSION`
+ * requires loading `civicrm.settings.php`, but that triggers a parse-error
+ * on PHP 5.x.)
+ *
+ * @see CRM_Upgrade_Form::MINIMUM_PHP_VERSION
+ * @see CiviDrupal\PhpVersionTest::testConstantMatch()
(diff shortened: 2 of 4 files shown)


totten:
I did some `r-run` on PHP 5.6 (web UI and CLI) to confirm they show the better PHP version error Similarly, I did `r-run` on PHP 7.0 (web UI and CLI) to confirm that they don't show a PHP version error.

totten:
jenkins, test this please

eileenmcnaughton:
Straight forward backport

PR #589, opened 2019-11-19
Fix issue $this

Dietermartens:
$this does not exist - changed to $contactId

### DIFF
--- a/modules/civicrm_engage/civicrm_engage.module
+++ b/modules/civicrm_engage/civicrm_engage.module
@@ -136,7 +136,7 @@ function civicrm_engage_civicrm_pageRun(&$page) {
     }
     $contactId = $page->getVar('_contactId');
     if (empty($contactId)) {
-      $contactId = CRM_Utils_Request::retrieve('cid', 'Positive', $contactId, TRUE, NULL, $_REQUEST);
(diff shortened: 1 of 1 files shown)


seamuslee001:
Jenkins ok to test

Dietermartens:
Maybe i should give some more info about this change. Its reproducible if you want to edit a birthdate on a contact view. When you click on save, it keeps loading infinitely. If you watch the ajax calls then you see an error on line 155 with a $this thats not defined.

modules/civicrm_engage/civicrm_engage.module
```diff
@@ -136,7 +136,7 @@ function civicrm_engage_civicrm_pageRun(&$page) {
     }
     $contactId = $page->getVar('_contactId');
     if (empty($contactId)) {
-      $contactId = CRM_Utils_Request::retrieve('cid', 'Positive', $this, TRUE, NULL, $_REQUEST);
+      $contactId = CRM_Utils_Request::retrieve('cid', 'Positive', $contactId, TRUE, NULL, $_REQUEST);
```
seamuslee001: Looking at the code here https://github.com/civicrm/civicrm-core/blob/master/CRM/Utils/Request.php#L74 i think the more correct option might be to use $page i think

Dietermartens: you're absolutely right!  I'll change that

PR #594, opened 2019-12-17
Update unit tests to use name spaced format of phpunit test classes f…

seamuslee001:
…or phpunit6 compatability


seamuslee001:
ping @eileenmcnaughton @totten

totten:
Change makes sense and passes `r-test`, which makes it good by definition.

PR #595, opened 2020-01-06
dev/drupal#102 Fix wiki link that goes nowhere

demeritcowboy:
https://lab.civicrm.org/dev/drupal/issues/102


seamuslee001:
looks good to me

PR #596, opened 2020-01-26
[NFC] Doc Block fix should be bool no boolean as per coding standards

seamuslee001:


totten:
Confirmed to pass stricter ruleset.

PR #599, opened 2020-03-11
Use new consolidated constant for minimum PHP version

agh1:
See https://github.com/civicrm/civicrm-core/pull/16753

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -35,12 +35,10 @@ define('CIVICRM_UF_HEAD', TRUE);
 /**
  * Minimum required PHP
  *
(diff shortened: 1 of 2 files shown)


agh1:
The regex was looking for x.y.z, and source data is x.y only.  I updated it to ignore the point release unless it's non-zero.

totten:
Looks good. :+1: Change makes sense. Tests pass. I did an `r-run` on `php70` and `php71` with `drush` (`drush ev 'civicrm_initialize(); echo "hi\n";'`), and (on `php70`) it complained correctly.

PR #601, opened 2020-03-29
provide Calendar module with end_date when duration is provided

KarinG:
Sharing this enhancement. It provides a very important enhancement for scheduling activities. 

**Before:** a CiviCRM calendar item only has a start_date so Drupal Calendar module only renders enough yellow block to ensure the text is encapsulated.

image

**After:** now constructing an end_date based on activity_duration -> if it exists in the View (regardless of whether it's displayed or not) use it. We now have a properly sized yellow block. The activity_duration in this example is 120 min and that is properly visualized:

image

### DIFF
--- a/modules/views/plugins/calendar_plugin_row_civicrm.inc
+++ b/modules/views/plugins/calendar_plugin_row_civicrm.inc
@@ -190,8 +190,19 @@ class calendar_plugin_row_civicrm extends calendar_plugin_row {
       }
       elseif (!$is_field && !empty($item)) {
         $item_start_date = new dateObject($item, $db_tz);
-        $item_end_date   = $item_start_date;
-        $node->date_id   = array('calendar.' . $node->id . '.' . $field_name . '.0');
+        $node->date_id = array('calendar.' . $node->id . '.' . $field_name . '.0');
+        // If we have a civicrm_activity_duration let's use it!
+        // Calculate the end_date so that the calendar item can have the correct height in the week and day displays.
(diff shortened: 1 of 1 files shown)


homotechsual:
(*CiviCRM Review Template WORD-1.2*)



* General standards
    * ([`r-explain`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-explain)) __PASS__
    * ([`r-user`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-user)) __PASS__: This would probably be the user-expected behaviour at present anyway!
    * ([`r-doc`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-doc)) __PASS__
    * ([`r-run`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-run)) __PASS__: Running on Master on D7 latest - no issues observed.
* Developer standards
    * ([`r-tech`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-tech)) __PASS__: No compatibility issues that I can tell - some possible edge cases if you've hacked your way around this in the past (which we have!) but I don't think we should factor people's "hacking" decisions into PR review!
    * ([`r-code`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-code)) __PASS__
    * ([`r-maint`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-maint)) __PASS__
    * ([`r-test`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-test)) __PASS__

homotechsual (approved):

PR #604, opened 2020-06-18
[REF] Use helper function to check if multiLingual

colemanw:
Overview
---------
Per the comment, this fixes a todo - using a helper function instead of doing it manually.

Before
---------
Duplicate code that is identical to helper function

After
----------
Just use the helper function.


colemanw:
@seamuslee001 this should be fine to merge now that the other PR is merged.

seamuslee001:
Makes sense merging

PR #606, opened 2020-07-06
Installation - Support "activate first" w/setup UI

totten:
Overview
----------

This adds support for installing CiviCRM-Drupal through the [Setup API](https://docs.civicrm.org/dev/en/latest/framework/setup/new-installer/). Most notable changes:

* Better layout and more useful options on the installation screen
* Extensible installation logic - under the hood, this installer can initialize settings, toggle extensions, etc.
* Reusable installation logic - can run the same installation code for CLI and web UI
* No need to bypass regular module activation or bypass router (no need for `install/index.php` entry-point)

Note: This is a port of https://github.com/civicrm/civicrm-backdrop/pull/121 to D7.

Before
------

The sysadmin downloads the CiviCRM module - but they must *not* install it. Instead, navigate to this path:

```
http://example.org/modules/civicrm/install/index.php
```

This is a special, standalone entry-point.

After
-----

After downloading the CiviCRM module, you simply enable it like any module. Then use any link to open a CiviCRM page (e.g. there are links in the post-install status-message and in the main menu). The page will display the setup screen.

For a sense of how it looks, here's a screenshot+mp4 from the Backdrop variant:

Screen Shot 2020-07-03 at 11 17 37 PM

http://think.hm/tmp/Backdrop-Civicrm-Setup.mp4

Technical Details
-----------------

1. The old install screen is still available as a fallback (in case there's some edge-case that got missed).
2. The basic gist of this change is:
    * Ensure that all CiviCRM-Drupal hooks fail gracefully (in the absence of `civicrm.settings.php`). Many already do - there were just a couple oddballs.
    * Update the handling of `civicrm/*` URLs. If CiviCRM cannot initialize, then show the [civicrm-setup web installer](https://docs.civicrm.org/dev/en/latest/framework/setup/new-installer/#web-installer-api).
3. This is the same API used in [the newer Civi-WP installer](https://github.com/civicrm/civicrm-wordpress/blob/5.28/civicrm.php#L1189-L1207). (*Note: For historical reasons, the Civi-WP has a few extra contingencies that aren't needed here.*)
4. When testing/`r-run`ning, here are some of things that I tried:
    * Enable the module and try to view it as regular/non-admin user. Regular users cannot perform installation.
    * Install via web UI 
    * Install via CLI (`cv core:install -vv --cms-base-url=http://dmaster.bknix:8001`), with the `civicrm.module` enabled or disabled
    * Install in US English or French.
4. During development, my workflow went a bit like this:
    1. Create D7 site with CiviCRM codebase -- but without installing. (*I specifically used a modified `civibuild` with the `drupal-demo` type - ie hack `app/config/demo-demo/install.sh` to skip most of the install logic. Then run `civibuild create dmaster` or `civibuild reinstall dmaster`*)
    2. Login.
    3. Make a new DB snapshot (eg `civibuild snapshot dmaster`). (This way my login-session is part of the snapshot.)
    4. Repeatedly do the following:
        * Edit code
        * Use the installer - and see how it goes.
        * Cleanup: `./bin/cv.phar core:uninstall -f ; civibuild restore dmaster ; rm -rf ~/bknix/build/dmaster/web/files/civicrm/`
5. This depends on another PR for civicrm-core: https://github.com/civicrm/civicrm-core/pull/17749

### DIFF
--- a/civicrm.install
+++ b/civicrm.install
@@ -36,6 +36,11 @@ function civicrm_enable() {
   menu_link_save($options);
 
   if (!civicrm_initialize()) {
+    $installLink = url('civicrm/setup');
+    $message = t('<p class="error"><strong>CiviCRM is almost ready.</strong> You must <a href="@url">configure CiviCRM</a> for it to work.</p>', [
+      '@url' => $installLink,
+    ]);
+    drupal_set_message($message);
     return;
   }
 
@@ -63,6 +68,27 @@ function civicrm_uninstall() {
 
   require_once 'CRM/Core/DAO.php';
   CRM_Core_DAO::dropAllTables();
+}
+
+function civicrm_requirements($phase) {
+  $requirements = array();
+  $t = get_t();
+  switch ($phase) {
+    case 'runtime':
+      if (!civicrm_initialize()) {
+        $installLink = url('civicrm/setup');
+        $requirements['civicrm'] = array(
+          'title' => $t("CiviCRM"),
+          'severity' => REQUIREMENT_ERROR,
+          'value' => $t('<p class="error"><strong>CiviCRM is almost ready.</strong> You must <a href="@url">configure CiviCRM</a> for it to work.</p>', [
+            '@url' => $installLink,
+          ]),
+        );
+        return $requirements;
+      }
+      break;
+  }
+  return $requirements;
 }
 
 /**
--- a/civicrm.install
+++ b/civicrm.install
@@ -65,35 +65,6 @@ function civicrm_uninstall() {
   CRM_Core_DAO::dropAllTables();
 }
 
-function civicrm_requirements($phase) {
-  global $base_url;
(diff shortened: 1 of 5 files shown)


seamuslee001:
I tested this and confirmed it works, I would note that this new setup routine doesn't allow you or at least not obvious to me how to install into a separate DB, it assumes you want to use the same DB as the CMS

totten:
> I tested this and confirmed it works, I would note that this new setup routine doesn't allow you or at least not obvious to me how to install into a separate DB, it assumes you want to use the same DB as the CMS

@seamuslee001 Good comment. Just to clarify - it should work with a separate DB, but it's an advanced option. When you skim the technical details at the bottom, the Civi DSN should have a pencil button. The styling is maybe too subtle.

Screen Shot 2020-07-15 at 7 31 50 PM

In any event, click that to edit.

Screen Shot 2020-07-15 at 7 32 06 PM

PR #607, opened 2020-07-06
fix deceased date format

yashodha:
fix deceased date format


eileenmcnaughton:
I don't think this module is used much now so happy to take your word on this

PR #608, opened 2020-07-13
fix url for file field

yashodha:


seamuslee001:
This just sets fragment to be the correct default of NULL merging

PR #610, opened 2020-07-29
NFC - Docblock cleanup

colemanw:
Cleanup code blocks


MegaphoneJon:
I just scrolled through all of this and agree that these are non-functional changes in the strictest sense of the term.

PR #611, opened 2020-08-05
dev/core#1929 Fix filter handler for custom field

aydun:
See https://lab.civicrm.org/dev/core/-/issues/1929

If the column_name is specified when the custom field is created, it does not contain the field id, but the handler currently assumes it does.

Change how the field id is determined and use the method currently in civicrm_handler_field_custom.inc


colemanw:
Less hacky :+1:

PR #612, opened 2020-08-11
Remove handling for 4.2 DBs

eileenmcnaughton:
I feel like we no longer support direct upgrades from 4.2 so we don't need this. There are some other
places in the file that could be done as follow ups

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -691,41 +691,16 @@ function drush_civicrm_upgrade_db_validate() {
  * Implementation of command 'civicrm-upgrade-db'
  */
 function drush_civicrm_upgrade_db() {
-  if (class_exists('CRM_Upgrade_Headless')) {
-    // Note: CRM_Upgrade_Headless introduced in 4.2 -- at the same time as class auto-loading
-    $codeVer = CRM_Utils_System::version();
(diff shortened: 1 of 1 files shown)


drush/civicrm.drush.inc
```diff
@@ -691,41 +691,16 @@ function drush_civicrm_upgrade_db_validate() {
...
-  }
+  $codeVer = CRM_Utils_System::version();
+  $dbVer = CRM_Core_BAO_Domain::version();
+  if (version_compare($codeVer, $dbVer) == 0) {
+    drush_print(dt('You are already upgraded to CiviCRM @version', ['@version' => $codeVer]));
+    return TRUE;
+  }
+  $upgradeHeadless = new CRM_Upgrade_Headless();
+  // FIXME Exception handling?
+  $result = $upgradeHeadless->run();
```
eileenmcnaughton, line 702: Tangentally - I wonder if we ensured Upgrade_Headless throws exceptions we could do a better job with drush / cv output. Currently the actual db error is lost

eileenmcnaughton:
@seamuslee001 @colemanw @mlutfy  - quite a lot of stuff in the drush file to support legacy upgrades - but we don't support pre 4.4 now do we (& even that only kinda)

seamuslee001:
Test fail unrelated merging

PR #615, opened 2020-08-21
Fixed for multi-select filter

pradpnayak:
In latest version of Civi multi-select field is flagged as serialize=1. Since this change was added to CiviCRM module, views failed to filter on multi-select field like multi-select, multi-select country and state. 

@colemanw


eileenmcnaughton:
@pradpnayak are you able to rebase this against the rc- I tried just switching it but it has other commits

pradpnayak:
@eileenmcnaughton rebased it

eileenmcnaughton:
Thanks @pradpnayak - I thought you would be in bed!

seamuslee001:
Tested it and agreed this is correct MOP

PR #316, opened 2015-09-15
CRM-16000 Member Role Sync: precedence on role grant/revoke

agh1:
Simply sort the memberships by status weight descending

---
- CRM-16000: Member Role Sync Uses Inactive Memberships Over Active Memberships
  https://issues.civicrm.org/jira/browse/CRM-16000

PR #623, opened 2020-09-09
Fix for #140, changing the logic to remove or add roles only if neede…

sadashivdalvi:
…d and always use google api instead of delete

### DIFF
--- a/modules/civicrm_member_roles/civicrm_member_roles.module
+++ b/modules/civicrm_member_roles/civicrm_member_roles.module
@@ -558,37 +558,6 @@ function _civicrm_member_roles_sync($ext_uid = NULL, $cid = NULL, $sync_type = N
       $error = $e->getMessage();
       return $error;
     }
-    //CRM-16000 remove inactive memberships if member has both active and inactive memberships
-    if ($memberships['count'] > 1) {
-      $params = array(
-        'sequential' => 1,
-        'name' => array('IN' => array("Deceased", "Cancelled", "Pending", "Expired")),
-      );
-      try {
-        $result = civicrm_api3('MembershipStatus', 'get', $params);
-      }
-      catch (CiviCRM_API3_Exception $e) {
-        $error = $e->getMessage();
-        return $error;
-      }
-      foreach ($result['values'] as $value) {
-        $inactive[] = $value['id'];
-      }
-      foreach ($memberships['values'] as $key => $membership) {
-        //Do not unset if inactive membership status is chosen as an option for synchronization.
-        $inactiveStatusSync = FALSE;
-        if (!empty($memberroles[$membership['membership_type_id']])) {
-          foreach ($memberroles[$membership['membership_type_id']] as $rolerule) {
-            if (in_array($membership['status_id'], $rolerule['codes']['current'])) {
-              $inactiveStatusSync = TRUE;
-            }
-          }
-        }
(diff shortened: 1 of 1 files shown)


petednz:
@jitendrapurohit can you look over this?

aydun:
@sadashivdalvi What is the problem you are addressing?  You mention '#140' but in which system?  The PR that is automatically linked does not look relevant.  I don't understand the comment 'always use google api instead of delete' ... the code makes no use of google api's.

sadashivdalvi:
I have a system which performs certain operation whenever a role is added or removed, this operation is to sync some external system.

What currently happens is, system removes all roles (using db delete which doesn't invoke any hooks, etc) and then adds roles which actually invokes the external system unnecessarily,  Instead of that civicrm code should identify whether we need a change in role and use api to save roles.

I am using this patch on production since I submitted the PR and it's working fine.

"google" was a typing mistake, I meant use api to save role changes.

Thanks,
Sadashiv

aydun:
Ok, that makes sense.  The changes are smaller than the diff suggests at first sight.  I have not tested it, but the code changes look good to me.
Let's see what Jitendra says.

modules/civicrm_member_roles/civicrm_member_roles.module
```diff
@@ -593,46 +593,44 @@ function _civicrm_member_roles_sync($ext_uid = NULL, $cid = NULL, $sync_type = N
     $contactMemberships = CRM_Utils_Array::value('values', $memberships);
     $addRoles           = array();
     $expRoles           = array();
-
-    $roleCondition = is_array($rid) ? 'IN' : '=';
-
-    if (empty($contactMemberships) && !empty($rid)) {
-      db_delete('users_roles')->condition('uid', $uid)->condition('rid', $rid, $roleCondition)->execute();
-    }
```
jitendrapurohit, line 601: is this no longer happening now? I just replicated the following problem which wasn't happening with the previous code.

- Create 2 member sync rule. Eg role1 => memtype1 & role2 => memtype2.
- Create a contact (with drupal user) and add a membership of type = memtype1.
- Sync the rule. role1 is added to the user (Correct).
- Delete the membership `memtype1` from the contact and add the membership of type = `memtype2`.
- Sync the rule again. role2 is added (Correct). But role1 is not removed from the contact (Incorrect).

As `memtype1` is not associated with the civi contact, role1 should have been removed after the sync. It was working as per the above condition.

@sadashivdalvi

sadashivdalvi: @jitendrapurohit can you please recheck, I missed that case, which I fixed by adding additional check now.

````
foreach ($memberroles as $membership_type_id => $rolerule) {
  if (user_has_role($rolerule['rid'], $account) && !in_array($rolerule['rid'], $addRoles)) {
    // This will only happen if the membership is deleted.
    $expRoles[] = $rolerule['rid'];
  }
}
````

Simple check that if the user has a role in the rule and doesn't have it addRoles then remove that role.

Have also improved the saving logic by moving the save out of loops, something I missed last time.

Thanks,
Sadashiv

sadashivdalvi (commented):
I did find on my site that we do need a fix from https://issues.civicrm.org/jira/browse/CRM-16000 i.e. https://github.com/civicrm/civicrm-drupal/pull/316  Should I include that patch here so that the code becomes complete and functional.  

I have this deployed on my production site.

Thanks,
Sadashiv

PR #624, opened 2020-09-16
Update docs links

homotechsual:
And replace forum link with SE. Sister PR to civicrm/civicrm-drupal-8#48

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -191,13 +191,13 @@ function civicrm_initialize() {
     // get ready for problems
     $docLinkInstall = "https://docs.civicrm.org/installation/en/latest/drupal7/";
     $docLinkTrouble = "https://docs.civicrm.org/installation/en/latest/general/troubleshooting/";
-    $forumLink      = "https://civicrm.stackexchange.com/questions/tagged/drupal";
+    $seLink      = "https://civicrm.stackexchange.com";
 
-    $errorMsgAdd = t("Please review the <a href='!1'>Drupal Installation Guide</a> and the <a href='!2'>Trouble-shooting page</a> for assistance. If you still need help installing, you can often find solutions to your issue by searching for the error message in the <a href='!3'>installation support section of the community forum</a>.</strong></p>",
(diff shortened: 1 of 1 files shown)


seamuslee001:
Jenkins add to whitelist

demeritcowboy:
This looks good I just have a similar comment to the one for drupal 8 that the reason down on line 216 where this is used isn't the most likely reason anymore. You don't even get this far if civicrm_root is incorrect. A possible reason to hit that line is that CRM/Core/Config.php is missing.

homotechsual:
> This looks good I just have a similar comment to the one for drupal 8 that the reason down on line 216 where this is used isn't the most likely reason anymore. You don't even get this far if civicrm_root is incorrect. A possible reason to hit that line is that CRM/Core/Config.php is missing.

I mean the check that throws this message is a check for CRM/Core/Config.php. Perhaps it's worth rephrasing the actual error message as well?

seamuslee001:
@MikeyMJCO this will need a rebase and its probably more about the language change in the message now as I think Eileen and I have fixed the links based on your feedback

homotechsual:
In which case I'm closing this out. I was only intending to fix the links, not correct the language.

homotechsual:
So I've just picked up what you mean about the language... Not sure why a PR was merged that changes a forum link to SE but leaves the text referencing the community forum??? Not sure why you didn't just merge this PR since it had the links changes AND the text changes...

And this PR wasn't touching the line @demeritcowboy was referencing anyway, so that comment shouldn't have blocked this PR.

homotechsual:
Rebased.

demeritcowboy:
@MikeyMJCO looks like the rebase didn't take?

And yes my comment here was just noting the parallel with the other PR.

homotechsual:
> @MikeyMJCO looks like the rebase didn't take?
> 
> And yes my comment here was just noting the parallel with the other PR.

I'm seeing no conflicts with base branch.

homotechsual:
Force pushed without the extra commit.

PR #628, opened 2020-11-18
Remove support for pre 4.2 Civi in drush

eileenmcnaughton:
Note there is still another plac ein the file with legacy version support


eileenmcnaughton:
@colemanw you like purging old versions...

PR #632, opened 2020-12-05
[NFC] - Remove literally empty if block

demeritcowboy:
The if block literally does nothing.


eileenmcnaughton:
You are right!

demeritcowboy:
Ah forgot about this. It's almost like it wasn't even there...

PR #634, opened 2020-12-09
allow filtering by empty event end, registration start, registration …

MegaphoneJon:
…end dates

Of the date fields on `civicrm_event` - start date, end date, registration start date, registration end date - only "start date" is required.  However, Views can't search for empty dates without explicitly specifying a date can be empty.  This adds that specification to the three event dates that aren't required.

### DIFF
--- a/modules/views/components/civicrm.event.inc
+++ b/modules/views/components/civicrm.event.inc
@@ -241,6 +241,7 @@ function _civicrm_event_data(&$data, $enabled) {
     'filter' => array(
       'handler' => 'civicrm_handler_filter_datetime',
       'is date' => TRUE,
+      'allow empty' => TRUE,
     ),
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
@jitendrapurohit @petednz - is this one you could review?

jitendrapurohit:
**BEFORE** -

No filter to specify date = empty.

image

**AFTER**

Can be filtered with date=empty

image

The condition also works fine in rendering the results -

>WHERE (( (civicrm_event.registration_start_date IS NULL ) AND (civicrm_event.end_date IS NOT NULL ) ))

jitendrapurohit (approved):

eileenmcnaughton:
Thanks for the speedy review @jitendrapurohit

PR #635, opened 2021-01-09
dev/core#2258 - Define CIVICRM_CRED_KEYS during drush installation

totten:
When using `drush civicrm-install`, the `civicrm.settings.php` should provide a credential-key.

See also: https://github.com/civicrm/civicrm-core/pull/19349 and https://lab.civicrm.org/dev/core/-/issues/2258

Here's an example usage:

```
## Remove old installation
[bknix-min:~/bknix/build/dmaster/web/sites/all/modules/civicrm/drupal] cv core:uninstall --db='mysql://dmasterciv_f6xwh:********@127.0.0.1:3307/dmastercivi_ssc8b?ne
w_link=true'
Found code for civicrm-core in /home/totten/bknix/build/dmaster/web/sites/all/modules/civicrm
Found code for civicrm-setup in /home/totten/bknix/build/dmaster/web/sites/all/modules/civicrm/setup
Found civicrm_* database tables in dmastercivi_ssc8b
Found civicrm.settings.php in /home/totten/bknix/build/dmaster/web/sites/default

Are you sure want to purge the CiviCRM database and data files? Data may be permanently destroyed. (y/N) y
Removing civicrm_* database tables in dmastercivi_ssc8b
Removing civicrm.settings.php from /home/totten/bknix/build/dmaster/web/sites/default

## Install Civi
[bknix-min:~/bknix/build/dmaster/web/sites/all/modules/civicrm/drupal] drush civicrm-install --site_url='dmaster.bknix:8001' --dbhost=127.0.0.01:3307 --dbname=dmastercivi_ssc8b --dbuser=dmasterciv_f6xwh --dbpass=********  --destination=sites/all/modules
CiviCRM database loaded successfully.                                                                                                                            [ok]
Settings file generated: /home/totten/bknix/build/dmaster/web/sites/default/civicrm.settings.php                                                                 [ok]
CiviCRM installed.                                                                                                                                               [ok]

## Work around quirk of drush installer
[bknix-min:~/bknix/build/dmaster/web/sites/all/modules/civicrm/drupal] drush ev 'civicrm_initialize();'

## Inspect the generated settings
[bknix-min:~/bknix/build/dmaster/web/sites/all/modules/civicrm/drupal] cv ev 'print_r(Civi::service("crypto.registry")->getKeys()); print_r([CIVICRM_CRED_KEYS]);'
Array
(
    [plain] => Array
        (
            [key] => 
            [suite] => plain
            [tags] => Array
                (
                )

            [id] => plain
[...]

PR #636, opened 2021-01-10
Define CIVICRM_CRED_KEYS during drush installation

seamuslee001:
Overview
----------------------------------------
When using `drush civivcrm-install`, the `civicrm.settings.php` should have a credential-key.

See also: civicrm/civicrm-core#19349 and https://lab.civicrm.org/dev/core/-/issues/2258

Before
----------------------------------------

Blank key

After
----------------------------------------

Random key


seamuslee001:
@totten

seamuslee001:
closing in favour #635 as identical

PR #638, opened 2021-02-12
dev/core#2364 Use civicrm_api3 and civicrm_api4 instead of civicrm_ap…

seamuslee001:
…i to better handle PEAR_Exceptions

ping @demeritcowboy @eileenmcnaughton

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -608,7 +608,7 @@ function drush_civicrm_ext_install($extension_name) {
     return;
   }
   try {
-    $result = civicrm_api('extension', 'install', ['key' => $extension_name, 'version' => 3]);
+    $result = civicrm_api3('extension', 'install', ['key' => $extension_name]);
     if ($result['values'] && $result['values'] == 1) {
       drush_print(dt('Extension !ename installed.', ['!ename' => $extension_name]));
     }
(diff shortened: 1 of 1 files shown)


seamuslee001:
(note that this would need to be ported to Backdrop but can handle once we are comfortable with this)

drush/civicrm.drush.inc
```diff
@@ -1577,9 +1578,31 @@ function drush_civicrm_api() {
...
+      }
+      catch (CiviCRM_API3_Exception $e) {
+        drush_set_error('CIVICRM api error', $e->getMessage());
+      }
+      break;
+
+    case 4:
+      try {
+        unset($params['version']);
+        $restul = civicrm_api4($entity, $action, $params);
```
demeritcowboy: typo: `restul`

seamuslee001: dang

drush/civicrm.drush.inc
```diff
@@ -1577,9 +1578,31 @@ function drush_civicrm_api() {
...
+      catch (CiviCRM_API3_Exception $e) {
+        drush_set_error('CIVICRM api error', $e->getMessage());
+      }
+      break;
+
+    case 4:
+      try {
+        unset($params['version']);
+        $result = civicrm_api4($entity, $action, $params);
+      }
```
demeritcowboy, line 1596: The PR is all good just noting that depending on the api call the return value from v4 doesn't print anything useful, so e.g. can't really be used for get operations. Long-term is it worth trying to support such a thing from drush?

eileenmcnaughton:
Looks good to me - if @demeritcowboy is happy then good to merge

demeritcowboy:
I'm happy. V4 is no different from before the patch just musing about future.

PR #644, opened 2021-03-30
Add FUNDING.yml to display a SPONSOR link that directs to c.o

homotechsual:


homotechsual:
See https://lab.civicrm.org/community/sustainability/-/issues/20

PR #646, opened 2021-07-07
[REF] Follow up fix to fixing note entity tables in views

seamuslee001:
ping @colemanw @totten @eileenmcnaughton

### DIFF
--- a/modules/views/components/civicrm.core.inc
+++ b/modules/views/components/civicrm.core.inc
@@ -1159,7 +1159,8 @@ function _civicrm_core_data(&$data, $enabled) {
   $data['civicrm_note']['table']['group'] = t('CiviCRM Note');
   $data['civicrm_note']['table']['join'] = array();
 
-  foreach (array_keys(CRM_Core_BAO_Note::entityTables()) as $entityTable) {
+  $noteEntities = civicrm_api3('Note', 'getoptions', ['field' => "entity_table"])['values'];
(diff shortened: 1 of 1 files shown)


modules/views/components/civicrm.core.inc
```diff
@@ -1159,7 +1159,8 @@ function _civicrm_core_data(&$data, $enabled) {
   $data['civicrm_note']['table']['group'] = t('CiviCRM Note');
   $data['civicrm_note']['table']['join'] = array();
 
-  foreach (array_keys(CRM_Core_BAO_Note::entityTables()) as $entityTable) {
+  $noteEntityTables = civicrm_api3('OptionValue', 'get', ['option_group_id' => 'note_used_for'])['values'];
+  foreach (array_values(CRM_Utils_Array::collect('value', $noteEntityTables)) as $entityTable) {
```
colemanw: I think both the **before** and **after** versions are incorrect because they access lower-level data instead of doing a pseudoconstant lookup and letting that layer worry about where the data is coming from. If it had been doing this from the beginning, we wouldn't need any change here:
```suggestion
  $noteEntities = civicrm_api3('Note', 'getoptions', ['field' => "entity_table"])['values'];
  foreach (array_keys($noteEntities) as $entityTable) {
```

seamuslee001:
good point @colemanw applied the patch here now

PR #647, opened 2021-07-23
dev/drupal#161 Update import of civicrm_generated.mysql to not use th…

seamuslee001:
…e lineMode as per core

This is as per https://github.com/civicrm/civicrm-core/pull/20925 @demeritcowboy @eileenmcnaughton

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -401,7 +401,7 @@ function _civicrm_install_db($dbuser, $dbpass, $dbhost, $dbname,
   civicrm_source($dsn, $acl_file);
 
   if ($loadGeneratedData) {
-    civicrm_source($dsn, $sqlPath . DIRECTORY_SEPARATOR . 'civicrm_generated.mysql', TRUE);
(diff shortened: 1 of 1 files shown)


demeritcowboy:
I'm going to disagree with this in the sense that as per https://github.com/civicrm/civicrm-core/pull/20925#issuecomment-884645689 it doesn't work with or without this PR and I don't think it ever did. I had posted in chat too but haven't got confirmation or denial yet. My vote would be to remove support for load_generated which brings it in line with the CLI for WP, Joomla, and even Drupal 8. My second-place vote would be to fix up the conditionals so it does work.

eileenmcnaughton:
I endorse removing code that doesn't work as a general principle

demeritcowboy:
Added a suggested note to dev-digest to see if anyone has ever had this working.

demeritcowboy:
Have added https://github.com/civicrm/civicrm-drupal/pull/648 as alternate since nobody has responded to the dev-digest saying they are using it.

demeritcowboy:
Closing since this line doesn't exist anymore.

PR #648, opened 2021-08-10
dev/drupal#161 - Remove drush sample data install option that doesn't work

demeritcowboy:
https://lab.civicrm.org/dev/drupal/-/issues/161

* It [doesn't work, probably never did, and it was even noted when it was added that it potentially had an issue](https://github.com/civicrm/civicrm-core/pull/20925#issuecomment-884645689)
* It was mentioned in the dev digest and nobody has responded that they are using it.
* None of the other CMS native CLI tools have this feature implemented for civi, including even Drupal 8.

The other option is to adjust it to make it work, but if nobody's using it is it needed? `cv` has an option to load the sample data if someone needs a CLI way to do it.

See also https://github.com/civicrm/civicrm-drupal/pull/647, but as-written it doesn't make it work.


eileenmcnaughton:
I'm sold - we did put it in the digest

PR #650, opened 2021-10-11
Fix unreplaced template vars in civicrm.settings.php when using drush installer

demeritcowboy:
See also https://github.com/civicrm/civicrm-joomla/pull/63


demeritcowboy:
Oh right this was still here. Thanks.

PR #652, opened 2022-01-14
civicrm.drush.inc - Add 'civicrm-pipe' subcommand

totten:
Add support for subcommand `drush civicrm-pipe` (`drush cvpipe`)

Complements https://github.com/civicrm/civicrm-core/pull/22262. Similar to https://github.com/civicrm/cv/pull/110.

Before
------------

Longer commands:

```bash
drush ev 'civicrm_initialize(); Civi::pipe();'
drush ev 'civicrm_initialize(); Civi::pipe("vlu");'
````

After
-------------

Shorter commands:

```bash
drush cvpipe
drush cvpipe vlu
```

Comment
--------------

If the command is successful, it will show a welcome/header line, e.g.

```javascript
{"Civi::pipe":{"v":"5.47.alpha1","l":["nologin"],"u":"untrusted"}}
```

You may then send JSON-RPC 2.0 requests, e.g.

```javascript
// Send request for `echo("hello world")`
{"jsonrpc":"2.0","method":"echo","params":["hello world"],"id":null}
// Receive response "hello world"
{"jsonrpc":"2.0","result":["hello world"],"id":null}
```

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -136,6 +136,18 @@ function civicrm_drush_command() {
   $items['civicrm-disable-debug'] = [
     'description' => "Disable CiviCRM Debugging.",
   ];
+  $items['civicrm-pipe'] = [
+    'description' => 'Start a Civi::pipe session (JSON-RPC 2.0)',
+    'examples' => [
+      'drush civicrm-pipe' => 'Begin a session with default flags',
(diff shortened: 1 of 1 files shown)


totten:
@seamuslee001 The test-output here looks OK (ie all `ok 123...` and `EXITCODES=""`). Is the failure something to do with the job-configuration?

seamuslee001:
Jenkins re test this please

PR #653, opened 2022-03-08
Views - CiviGrant is now an extension

colemanw:
As of 5.47, CiviGrant is an Extension, not a Component.

PR #654, opened 2022-03-08
Views - CiviGrant is now an extension

colemanw:
PR for 5.48
See #653

### DIFF
--- a/modules/views/civicrm.views.inc
+++ b/modules/views/civicrm.views.inc
@@ -67,6 +67,11 @@ function civicrm_views_data_alter(&$data) {
 
   // Get list of enabled CiviCRM components
   $enabled = CRM_Core_Component::getEnabledComponents();
+  // CiviGrant is now an extension
+  $civiGrant = civicrm_api3('Extension', 'get', [
+    'key' => 'civigrant',
+    'status' => 'installed',
+  ]);
 
   // $data = array();
   // Load Core CiviCRM data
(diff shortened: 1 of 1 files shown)


seamuslee001:
Jenkins retest this please

demeritcowboy:
I had some issues with civicrm_queue after the patch but I think they're unrelated. Will sort that out later. But this doesn't seem to fix the problem but I don't know why yet.

BEFORE
---------
1. Set up drupal with views (single-db so don't have to deal with that).
2. Enable civigrant component.
3. Make some grants.
4. Make a view that shows some grants.
5. Upgrade - progress bar hangs. Error class CRM_Grant_DAO_Grant not found.
6. Grant extension appears disabled.

AFTER
-------
5. Upgrade - progress bar hangs. Error class CRM_Grant_DAO_Grant not found.
6. Grant extension appears disabled.

colemanw:
Thanks @demeritcowboy this PR isn't actually intended to address that upgrade problem, but to allow views of grants to continue working normally, post-upgrade.

I'm trying to recreate the upgrade problem locally and if I manage to do that then I'll work through it with a debugger.

demeritcowboy:
Ok I think I see what my problem is. If I test things like a dev instead of a sysadmin then it upgrades fine, but either way not related to this PR. I'll write that up in a ticket later since it might be more of a thing now.

PR #656, opened 2022-03-10
civicrm_handler_filter_pseudo_constant - Guard against old or inactive pseudoconstants

totten:
Example 1: You install an extension which defines some data (pseudo-constant).  A view depends on that data.  You remove the extension.  It shouldn't crash.

Example 2: Upgrading to 5.47.x with CiviGrant installed.  You have some views that rely on CiviGrant.  During upgrade/migration, it momentarily disables CiviGrant (component) -- and then later re-enables CiviGrant (extension).  If there's a system-flush during the interim, the CiviGrant data isn't available yet. Don't crash.

Note: I haven't tested this patch. It's a draft based on reading error-report/backtrace.

### DIFF
--- a/modules/views/civicrm/civicrm_handler_filter_pseudo_constant.inc
+++ b/modules/views/civicrm/civicrm_handler_filter_pseudo_constant.inc
@@ -40,8 +40,15 @@ class civicrm_handler_filter_pseudo_constant extends views_handler_filter_in_ope
       $pseudo_args = array();
     }
 
-    // Include and call the Pseudo Class method
-    $this->_pseudo_constant = call_user_func_array($this->definition['pseudo class'] . "::" . $this->definition['pseudo method'], $pseudo_args);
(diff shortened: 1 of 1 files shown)


colemanw:
Code looks good. If it passes testing, will need backporting to 5.47

colemanw:
Merging per `r-run`

seamuslee001:
pinging @eileenmcnaughton

demeritcowboy:
I think either this has come back in some other scenario, or it maybe was never a complete solution. There's also code in civicrm_handler_`field`_pseudo_constant that calls the same thing, but note the $callback is `CRM_Core_PseudoConstant::get`, which is always callable even when the $pseudo_args point to something that isn't, in this case
```
Array
(
    [0] => CRM_Grant_DAO_Grant
    [1] => grant_type_id
)
```

PR #666, opened 2023-05-18
Add more error output for db error in drush

eileenmcnaughton:
Without this it just says there was an error - no clue as to what

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -27,6 +27,8 @@
  *     3. In an arbitrary folder specified with the --include option.
  */
 
+use Civi\Core\Exception\DBQueryException;
+
 /**
  * Implements hook_drush_command().
  *
@@ -1605,7 +1607,12 @@ function drush_civicrm_api() {
         unset($params['version']);
         $result = civicrm_api4($entity, $action, $params);
       }
-      catch (API_Exception $e) {
+      catch (DBQueryException $e) {
+        drush_set_error('CIVICRM api error', $e->getDBErrorMessage() . ' ' . $e->getUserMessage()
+          . (in_array('debug=1', $args, TRUE) ? "\n" . $e->getSQL() : '')
+        );
+      }
+      catch (CRM_Core_Exception $e) {
         drush_set_error('CIVICRM api error', $e->getMessage());
       }
       break;


drush/civicrm.drush.inc
```diff
@@ -1605,7 +1607,12 @@ function drush_civicrm_api() {
         unset($params['version']);
         $result = civicrm_api4($entity, $action, $params);
       }
-      catch (API_Exception $e) {
+      catch (DBQueryException $e) {
+        drush_set_error('CIVICRM api error', $e->getDBErrorMessage() . ' ' . $e->getUserMessage()
+          . (in_array('debug=1', $args, TRUE) ? "\n" . $e->getSQL() : '')
```
demeritcowboy, line 1612:
```suggestion
          . (!empty($params['debug']) ? ("\n" . $e->getSQL()) : '')
```
$args has already been parsed and put into params, and depending on how you specify the args it might not look like `debug=1`. In fact I couldn't even figure out how to use this for api4 without using the json method where it's not in args.

Also for me getSQL() returned blank, but that's not a blocker maybe it depends what the error situation is. For me I forced a db syntax error.

I also get a secondary error `Undefined variable $result`, but it's not because of the PR. A followup could wrap the output block below to only run if `$result` actually got set.

eileenmcnaughton: @demeritcowboy then I think maybe checking debug is too complex & just add the extra output?

eileenmcnaughton:
I just trialled this patch as getting the data better - of course now I found why I had an error I need to switch to solving it

```
+use Civi\Core\Exception\DBQueryException;
+
 /**
  * Implements hook_drush_command().
  *
@@ -1592,8 +1594,14 @@ function drush_civicrm_api() {
         unset($params['version']);
         $result = civicrm_api3($entity, $action, $params);
       }
-      catch (CiviCRM_API3_Exception $e) {
+      catch (CRM_Core_Exception $e) {
         drush_set_error('CIVICRM api error', $e->getMessage());
+        drush_print('CIVICRM api error', $e->getMessage());
+        $nestedException = $e->getErrorData()['exception'] ?? NULL;
+        if ($nestedException && is_a($nestedException, '\Civi\Core\Exception\DBQueryException')) {
+          drush_set_error('CIVICRM api error', $nestedException->getDBErrorMessage() . ' ' . $nestedException->getUserMessage() . $nestedException->getSQL() ? "\n" . $nestedException->getSQL() : '');
+          drush_print('CIVICRM api error', $nestedException->getDBErrorMessage() . ' ' . $nestedException->getUserMessage() . $nestedException->getSQL() ? "\n" . $nestedException->getSQL() : '');
+        }
       }
```

PR #667, opened 2023-06-07
Fixed notice error defined CIVICRM_SETTINGS_PATH

pradpnayak:
When running ```drush cc civicrm``` we get ````Constant CIVICRM_SETTINGS_PATH already defined civicrm.drush.inc:1512````

### DIFF
--- a/drush/civicrm.drush.inc
+++ b/drush/civicrm.drush.inc
@@ -1494,25 +1494,31 @@ function _civicrm_init($fail = TRUE, $load_config = TRUE) {
     ]));
   }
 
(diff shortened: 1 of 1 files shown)


demeritcowboy:
Hi @pradpnayak can you give reproduction steps? I tried both plain and with doing `export CIVICRM_SETTINGS_PATH=/path/to/civicrm.settings.php` first and don't see any errors.

demeritcowboy:
@pradpnayak Is this solved by https://github.com/civicrm/civicrm-drupal/pull/668 ?

PR #668, opened 2023-07-27
Use full setting path rather than relative

eileenmcnaughton:
This makes the setting of CIVICRM_SETTINGS_PATH consistent with other places and addresses a build bug

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -179,9 +179,10 @@ function civicrm_initialize() {
 
   if (!$initialized) {
     if (function_exists('conf_path')) {
-      $settingsFile = conf_path() . '/civicrm.settings.php';
+      $settingsFile = DRUPAL_ROOT . '/' . conf_path() . '/civicrm.settings.php';
     }
     else {
+      // @todo - ensure this is not providing a relative path
       $settingsFile = conf_init() . '/civicrm.settings.php';
     }
     if (!defined('CIVICRM_SETTINGS_PATH')) {


totten:
I'm not really sure if absolute or relative is better, but consistency sounds good. With this patch, the following boot protocols would be in agreement:

* CMS-first boot for Drupal (*PR changes it to absolute*)
* CMS-first boot for WP (*already uses absolute*)
* Civi-first boot for all UFs (*already uses absolute*)

eileenmcnaughton:
Just FYI - I was looking at consistency within drupal - ie it is set to full path elsewhere on my drupal install - eg. the drush file, civicrm.config.php, Preboot.civi.setup.php

totten:
On `r-tech`nical impact -- I think this is OK, though it's theoretically possible for some customization/integration to be affected. The important thing is that the contract is still basically compatible.

* Yesterday, a robust consumer would need to allow for relative or absolute values.
* Today, a robust consumer should still allow for relative or absolute values. (*We haven't thoroughly combed through all possible environments...*)
* Tomorrow (or 4 months from now... or somesuch...), as it gets more consistent, consumers can start to assume that values are absolute.

demeritcowboy:
Is this the same issue as https://github.com/civicrm/civicrm-drupal/pull/667?

totten:
They sound different to me, eg

* In 667, it sounds like CIVICRM_SETTINGS_PATH is defined twice (in the same use-case/process/request).
* In 668, it sounds like CIVICRM_SETTINGS_PATH is defined once - but the value can be different (depending on the use-case/process/request)

PR #669, opened 2023-09-07
Views notice fix. Remove duplicate call to constructor.

demeritcowboy:
Minor followup to https://github.com/civicrm/civicrm-core/pull/27358

Look down a few lines where it defaults to array, and then down a few more lines where it's expecting an array to loop over.


demeritcowboy:
jenkins retest this please (although looks like maybe outdated tooling?)

seamuslee001:
Jenkins re test this please

PR #674, opened 2023-10-06
Fix Deprecated  string interpolation issue in civicrm.module (line 10…

awasson:
Resolves [#4694](https://lab.civicrm.org/dev/core/-/issues/4694)

Fixes Deprecated ${} string interpolation issue in civicrm.module (line 1058 of /civicrm/drupal/civicrm.module).

In PHP 8.2 ${} string interpolation has been deprecated:
https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation

> echo "Hello ${name}"; // Deprecated
> echo "Hello {$name}"; // Supported

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -1055,7 +1055,7 @@ function _civicrm_filter_process($text, $filter, $format, $langcode, $cache, $ca
 
   $was_secure       = $smarty->security;
   $smarty->security = TRUE;
-  $text             = $smarty->fetch("string:${text}");
+  $text             = $smarty->fetch("string:{$text}");
   $smarty->security = $was_secure;
 
   // In the use-case of embedding Smarty codes inside a Drupal page, one is likely


demeritcowboy:
This is obviously correct but I'm curious how do you enable this filter? I didn't even know it existed. And I don't see it in the list of text formats?

awasson:
I'm not entirely certain how this one is enabled but this particular site that runs on PHP 8.2 is integrated with the CiviCRM Webform module for membership applications so it might be that the filter is enabled when pulling CiviCRM contact data into a Webform fields on the Drupal side or when webform data is saved to CiviCRM on that side.

demeritcowboy:
Oh wow. Instructions are here: https://issues.civicrm.org/jira/browse/CRM-8817. When you go to manually add a text format it's one of the choices for "Enabled filters".

PR #675, opened 2023-10-06
dev/core#4695 Fix Deprecated function: Optional parameter declared before required

awasson:
Resolves [#4695](https://lab.civicrm.org/dev/core/-/issues/4695)

In PHP 8.2 declaring an optional parameter before a required parameter has been deprecated.
https://php.watch/versions/8.0/deprecate-required-param-after-optional  

The sites that this issue impacts use the Drupal Rules module in their workflow so this may be an issue that is only specific to sites that use Drupal Rules. I don't know if that is the case for certain.

### DIFF
--- a/modules/civicrm_rules/civicrm_rules.contact-eval.inc
+++ b/modules/civicrm_rules/civicrm_rules.contact-eval.inc
@@ -24,7 +24,7 @@
  */
 require_once 'civicrm_rules_utils.inc';
 
-function civicrm_rules_rules_action_contact_send_email($to, $subject, $message, $from = NULL, $settings, RulesState $state, RulesPlugin $element) {
+function civicrm_rules_rules_action_contact_send_email($to, $subject, $message, string $from = NULL, $settings, RulesState $state, RulesPlugin $element) {
 
   $to = str_replace(array("\r", "\n"), ',', $to);
   $toEmails = explode(',', $to);
--- a/modules/civicrm_rules/civicrm_rules.contact-eval.inc
+++ b/modules/civicrm_rules/civicrm_rules.contact-eval.inc
@@ -24,7 +24,7 @@
  */
 require_once 'civicrm_rules_utils.inc';
 
-function civicrm_rules_rules_action_contact_send_email($to, $subject, $message, string $from = NULL, $settings, RulesState $state, RulesPlugin $element) {
+function civicrm_rules_rules_action_contact_send_email($to, $subject, $message, $from, $settings, RulesState $state, RulesPlugin $element) {
 
   $to = str_replace(array("\r", "\n"), ',', $to);
   $toEmails = explode(',', $to);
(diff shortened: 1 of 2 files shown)


modules/civicrm_rules/civicrm_rules.contact-eval.inc
```diff
@@ -24,7 +24,7 @@
  */
 require_once 'civicrm_rules_utils.inc';
 
-function civicrm_rules_rules_action_contact_send_email($to, $subject, $message, $from = NULL, $settings, RulesState $state, RulesPlugin $element) {
+function civicrm_rules_rules_action_contact_send_email($to, $subject, $message, string $from = NULL, $settings, RulesState $state, RulesPlugin $element) {
```
eileenmcnaughton: this typing should be `?string` shouldn't it?

Also don't you need to remove the `=NULL`

ie the signature should be this 

`function civicrm_rules_rules_action_contact_send_email($to, $subject, $message, $from, $settings, RulesState $state, RulesPlugin $element) {`

OR  (more correct but slightly riskier) 


`function civicrm_rules_rules_action_contact_send_email($to, $subject, $message, ?string $from, $settings, RulesState $state, RulesPlugin $element) {`

awasson: Hi Eileen,

I've done some reading up on this issue and it's a little cloudy but I believe that the consensus is: 
`string $from = NULL` will not throw an error 
`?string  $from = NULL` will mostly not throw an error but technically still could.

I chose to go the safer route. 

Cite: https://externals.io/message/108052#108410 

I thought that I might simply delete ` = NULL` but I didn't want to do that in case there is a specific reason it was declared that way. Logically, there is no way to call a function with an empty optional argument followed by a required argument but I didn't want to second guess what the original developer had in mind. 

Cheers,
Andrew

eileenmcnaughton: @awasson so there are 2 things here

1) type hints
2) Optional parameter declared before required

The latter is what is causing you a problem & the former is what you are fixing - in general type hints are good coding but when they meet bad coding you get errors. So good to add, but has risks, and on code that is on it's way out the risks may not be work taking

On the other hand there is no risk in fixing the actual error - ie ' Optional parameter declared before required'

- this is occurring because `$from` is an optional variable. It is optional because the `= NULL` means that the calling function does not have to pass it in & if it does not NULL will be used. However, `$settings`, `$state` and `$element` are required - there is no default value. It is impossible to not pass a variable for `$from` because even though it is optional the following variables are required.

Hence the fix to your issue is to remove the `= NULL` from after the from.

awasson: That sounds like a solid solution. 

I'm not a frequent Pull Requester. If I fix that and push to my branch, do I make another pull request or can it merge with this one?

Thanks,
Andrew

eileenmcnaughton: @awasson the short answer is that in this case what you have done is fine & if it passes I will merge it

The long answer is that we prefer you to squash your commits & there are different ways to do that - usually involving `rebase -i` and a force push `git push -f`. However, given that this code is kinda end of life & not really being looked at I'm just gonna go with '2 commits to do 1 thing is fine this time'

PR #681, opened 2024-05-06
Replace db_delete with user_save for consistency in process

yogesh-compuco:
Replace db_delete with user_save for consistency in process

- When membership status gets updated in civicrm, civicrm_member_roles module update user with role change using user_save() function.
- Same process not followed for membership status gets deleted.
- We use db_delete() function to delete user role from that specific user.
- We face this issue when we are syncing user status with cognito & found out that in this case no user hooks gets executed like hook_user_update() or hook_user_presave().
- For consistency we need to utilise same user_save() function so functionalities dependent on this will keep intact.

### DIFF
--- a/modules/civicrm_member_roles/civicrm_member_roles.module
+++ b/modules/civicrm_member_roles/civicrm_member_roles.module
@@ -597,7 +597,12 @@ function _civicrm_member_roles_sync($ext_uid = NULL, $cid = NULL, $sync_type = N
     $roleCondition = is_array($rid) ? 'IN' : '=';
 
     if (empty($contactMemberships) && !empty($rid)) {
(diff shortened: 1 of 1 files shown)


demeritcowboy:
jenkins test this please

demeritcowboy:
Test fail is a known issue - ignoring.

Putting merge-ready since this looks the same as what happens down on line 625. (EDIT: There is no merge-ready label in this repo, so just pretend there is.)

demeritcowboy:
@yogesh-compuco For release notes credit can you do a PR in the civicrm-core repo to add yourself to https://github.com/civicrm/civicrm-core/blob/master/contributor-key.yml ?

yogesh-compuco:
Thanks @demeritcowboy , as suggested I have created a PR - https://github.com/civicrm/civicrm-core/pull/30149

PR #682, opened 2024-07-31
REF - Remove unnecessary uses of CRM_Utils_Array::value

colemanw:
Just fixing up some old code to stop using a deprecated function

### DIFF
--- a/modules/civicrm_engage/Engage/Report/Form/WalkList.php
+++ b/modules/civicrm_engage/Engage/Report/Form/WalkList.php
@@ -240,10 +240,10 @@ class Engage_Report_Form_WalkList extends Engage_Report_Form_List {
           //echo "&nbsp;&nbsp;&nbsp;field name $fieldName<br>";
           $clause = NULL;
 
-          if (CRM_Utils_Array::value('type', $field) & CRM_Utils_Type::T_DATE) {
-            $relative = CRM_Utils_Array::value("{$fieldName}_relative", $this->_params);
-            $from     = CRM_Utils_Array::value("{$fieldName}_from", $this->_params);
-            $to       = CRM_Utils_Array::value("{$fieldName}_to", $this->_params);
+          if (($field['type'] ?? 0) & CRM_Utils_Type::T_DATE) {
+            $relative = $this->_params["{$fieldName}_relative"] ?? NULL;
+            $from = $this->_params["{$fieldName}_from"] ?? NULL;
+            $to = $this->_params["{$fieldName}_to"] ?? NULL;
 
             $clause = $this->dateClause($field['name'], $relative, $from, $to);
           }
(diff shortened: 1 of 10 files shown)


demeritcowboy:
Test has failed twice but I can't see it being from anything here. Let's try one more time.

modules/civicrm_engage/Engage/Report/Form/WalkList.php
```diff
@@ -418,10 +418,7 @@ public function executePrintmode($rows) {
 
       $dob  = $value['civicrm_contact_birth_date'];
       $age  = empty($dob) ? 0 : $this->dob2age($dob);
-      if (!empty($value['civicrm_contact_gender_id'])) {
-        $sex = $gender[CRM_Utils_Array::value('civicrm_contact_gender_id', $value)];
-      }
-      $sex  = empty($sex) ? '' : $sex;
+      $sex = $gender[$value['civicrm_contact_gender_id']] ?? '';
```
demeritcowboy, line 421: Technically this could give a warning because the `$value['civicrm_contact_gender_id']` is evaluated first, but gender is (currently) a required field so this works out.

demeritcowboy:
```
api\v4\Afform\AfformConditionalUsageTest::testConditionalRequiredFieldBoolInput
array_key_exists(): The first argument should be either a string or an integer

/home/homer/buildkit/build/build-4/web/sites/all/modules/civicrm/ext/afform/core/Civi/Afform/Utils.php:43
/home/homer/buildkit/build/build-4/web/sites/all/modules/civicrm/ext/afform/core/Civi/Api4/Action/Afform/AbstractProcessor.php:522
/home/homer/buildkit/build/build-4/web/sites/all/modules/civicrm/ext/afform/core/Civi/Api4/Action/Afform/Submit.php:85
/home/homer/buildkit/build/build-4/web/sites/all/modules/civicrm/ext/afform/core/Civi/Api4/Action/Afform/AbstractProcessor.php:80
/home/homer/buildkit/build/build-4/web/sites/all/modules/civicrm/Civi/Api4/Provider/ActionObjectProvider.php:70
/home/homer/buildkit/build/build-4/web/sites/all/modules/civicrm/Civi/API/Kernel.php:153
/home/homer/buildkit/build/build-4/web/sites/all/modules/civicrm/Civi/Api4/Generic/AbstractAction.php:256
/home/homer/buildkit/build/build-4/web/sites/all/modules/civicrm/ext/afform/mock/tests/phpunit/api/v4/Afform/AfformConditionalUsageTest.php:137
```

Theories about why the test is failing:

* It could be that test has a problem on php 7.4. I can't find where the core ext tests regularly run on 7.4 so can't compare.
  * (Running a manual run here: https://test.civicrm.org/job/CiviCRM-Manual-Test/161/)
* Something about the test environment here is maybe out of date? It's been a while since tests ran here so can't compare.

In any case the PR runs should probably be updated to a newer php, whether it's related or not.

demeritcowboy:
Hmm ok that test does fail on php 7.4: https://test.civicrm.org/job/CiviCRM-Manual-Test/161 but not on php 8: https://test.civicrm.org/job/CiviCRM-Manual-Test/162

So will merge here and leave that as a separate issue.

Filed an infra/ops ticket for the environment here: https://lab.civicrm.org/infra/ops/-/issues/1024

modules/civicrm_engage/Engage/Report/Form/WalkList.php
colemanw: Maybe? This is not documented in the php docs as far as I can tell, but it was my understanding that the operator would guard against *anything* on the left side being null, but I could be wrong.

demeritcowboy: I checked with "php -r". It does guard e.g. $x[$y][2], but not $x[$y[2]]

PR #683, opened 2024-08-14
Update font awesome URL for installer page to V6

ufundo:
Following https://github.com/civicrm/civicrm-core/pull/30779 the URL for font awesome css file has changed, which means the tick on the install confirmation screen is missing.


demeritcowboy:
I don't see a tick with or without this, but I can see a 404 on the original url so this fixes that so going to merge.

The test fails have come up before and are because there's some afform code that doesn't work on php 7.4, just not sure if it's just test code or real code. See also https://lab.civicrm.org/infra/ops/-/issues/1024.

PR #688, opened 2024-09-10
D7 Views - Update some deprecated pseudoconstant function calls

colemanw:
Just in time for the D7 EOL announcement...


demeritcowboy:
`#drupal74eva`

If MySpace can still be in the list of website types, then drupal 7 is not even close to dead.

PR #691, opened 2025-07-03
dev/core#5951 Fetch contact_id if not present

eileenmcnaughton:
Addresses the possibility that contact_id may not be passed to hook


demeritcowboy:
From the ticket this looks like a theoretical PR but it seems harmless.

PR #692, opened 2025-09-26
Remove Unnecessary Loading Of Themes For Mailing Urls

shahrukh-compuco:
## Overview
This PR restricts the loading of civicrm themes for mailing open and mailing url tracking events as these does not require the theme loading and unnecessarily increases the server load which can bring entire site down if the mailing list is big.


ufundo:
This seems good - is it something that could be implemented in `civicrm-core` ?

shahrukh-compuco:
> This seems good - is it something that could be implemented in `civicrm-core` ?

I dont think so as its about drupal theme not civicrm theme

PR #695, opened 2026-01-22
Switch to using non deprecated class references in civicrm_api drush …

seamuslee001:
…function and return out of the function after setting error using drush_set_error to avoid notice about undefined variable

ping @colemanw @djnatedagreat


colemanw:
Looks like a no-brainer.

